PR #4, opened 2021-11-16
Add loading indicator

cblage:
This adds a loading indicator that gets displayed/hidden whenever an API call is made.

Also upgraded the nodejs version in `.tool-versions` to be compatible with the Apple M1 processor.


packages/taraxa-ui/package.json
```diff
@@ -32,32 +32,33 @@
     "@material-ui/core": "^4.12.2",
```
cblage, line 32: yarn reordered these when I added `clsx` as a dependency

PR #17, opened 2022-01-04
Render Trophy icon on top nodes

farebord:
This PR adds the Trophy icon to `@taraxa-ui` and renders it on top nodes in the delegation page.

### DIFF
--- a/services/community-site/src/pages/Delegation/Delegation.tsx
+++ b/services/community-site/src/pages/Delegation/Delegation.tsx
@@ -8,6 +8,7 @@ import {
   Switch,
   Text,
   Icons,
+  Card,
 } from '@taraxa_project/taraxa-ui';
 
 import {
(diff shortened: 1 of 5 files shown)


services/community-site/src/pages/Delegation/Delegation.tsx
```diff
@@ -334,7 +345,16 @@ const Delegation = () => {
                   delegatableNodes.map((row) => (
                     <TableRow className="tableRow">
                       <TableCell className="tableCell">{row.status}</TableCell>
-                      <TableCell className="tableCell">{row.name}</TableCell>
+                      <TableCell className="tableCell">
+                        <div className="flexCell">
```
rattrap, line 349: I think the icon is an inline element so no need to add it in a div to make it a block element and then to add flex over the name and icon

rattrap (changes requested):

services/community-site/src/pages/Delegation/Delegation.tsx
farebord: As I said on Slack:
```
My idea of adding a flexbox is just to add space-between the trophy icon and the title of the node. I’ve added the legend.
Using flex is probably the easier thing here since there is some space between both elements (title and icon). Do you agree or could you suggest other approaches?
```
imagen

PR #18, opened 2022-01-11
Use server side isActive flag, fix calculation of total blocks produced in run a node section

cblage:


services/community-site/src/pages/Delegation/Delegation.tsx
```diff
@@ -35,15 +35,14 @@ interface Node {
   user: number;
   name: string;
```
rattrap, line 36: Can you move this interface to a new file so we can use it in the `EditNode` screen also?

services/community-site/src/pages/RunNode/Screen/EditNode.tsx
```diff
@@ -11,7 +11,7 @@ interface Node {
   name: string;
```
rattrap, line 11: This interface should be in a separate file.

rattrap (changes requested):

rattrap (approved):

PR #21, opened 2022-01-17
[HOTFIX] get isTopNode directly from node instead of deconstructing 

farebord:
This PR removes partial deconstruction in favor of accessing `isTopNode` attribute directly from the node.


rattrap:
That wasn't the problem. We had two tables. One for regular nodes and one for fully delegated nodes. The icon wasn't displayed in the second one. Fixed in - https://github.com/Taraxa-project/taraxa-platform/commit/2319d24837bf6e4228ae3e1a7b221fd11f80a90a

PR #92, opened 2022-03-08
Implement table with mocked data

VargaElod23:


farebord:
Is there a reason we use `moment` instead of the default [Date](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date) object?

VargaElod23:
Out of personal convenience, honestly. I've been working with `moment` for a long time now and my personal experience is that developers that manipulate the code later on do it so much easier because moment's syntax is much more readable that  native Date object manipulation...

services/community-site/src/pages/Redeem/redeem.scss
```diff
@@ -1,3 +1,15 @@
+.container {
```
rattrap, line 1: You need to add these in a `.redeem` wrapper or something. Or else all of these classes will be applied to everywhere.

services/community-site/src/utils/claim.ts
```diff
@@ -0,0 +1,80 @@
+import { ethers } from 'ethers';
```
rattrap, line 1: I wouldn't add it to utils. Maybe create a new service for this?

rattrap (changes requested):

services/community-site/src/pages/Redeem/redeem.scss
Elod23: I'm sorry but I don't really get what you mean by this. Aren't these classes only on the scope of the JSX file it gets imported into?

services/community-site/src/utils/claim.ts
VargaElod23: Done 👍🏼

PR #102, opened 2022-03-22
Improvement/general start script

VargaElod23:


VargaElod23:
Follows #60 . Please don't merge until

VargaElod23:
@rattrap  the only item missing now from the full setup is proper env setup in `/services/delegation/.env.dist` and then we're there 👍🏼

PR #118, opened 2022-03-28
Updated profile on run a node page

### LINKED ISSUE #46: Feature: Community site - Delegation - Update "Create Profile"/"Update Profile" box
Update the "Create Profile" and "Update Profile" boxes on the "Run a node section" of the community site to reflect the new design in Figma.

https://community.taraxa.io/node

Profile exists:
image

Profile doesn't exist:
image

farebord:
Fixed #46

### DIFF
--- a/services/community-site/src/pages/RunNode/RunNode.tsx
+++ b/services/community-site/src/pages/RunNode/RunNode.tsx
@@ -191,6 +191,21 @@ const RunNode = () => {
   }, [isLoggedIn]);
 
   useEffect(() => {
+    delegationApi
+      .get(`/nodes?type=testnet`, true)
+      .then((r) => {
+        if (r.success) setTestnetNodes(r.response);
+      })
+      .catch(() => setTestnetNodes([]));
+    delegationApi
+      .get(`/nodes?type=mainnet`, true)
+      .then((r) => {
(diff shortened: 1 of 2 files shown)


services/community-site/src/pages/RunNode/RunNode.tsx
```diff
@@ -195,6 +199,13 @@ const RunNode = () => {
       return;
     }
 
+    const testnetNodesRequest = await delegationApi.get(`/nodes?type=testnet`, true);
```
VargaElod23, line 202: This is a bit confusing since you don't reuse it later on, wouldn't a simple `useEffect((), [])` be more vocal about the use-case? Tho if you wanna leave the implementation like this these events can be also fired when the user toggles between the views, which would be a nice feature. 🌮

VargaElod23 (approved):
LGTM, small question added.

services/community-site/src/pages/RunNode/RunNode.tsx
farebord: 👍

farebord: Updated this to be called in another useEffect.

Elod23 (approved):
LGTM. 🥳🎅🏼

PR #123, opened 2022-04-04
#91: Added styled components to community-site

### LINKED ISSUE #91: Improvement: Community site - Use styled components instead of SCSS
It would be great to use the same styling mechanism across the whole ecosystem, therefore it'd be great to use styled components not only in the Taraxa UI but in the community site too.

farebord:
Fixes #91


VargaElod23 (approved):
LGTM 👍🏼

PR #145, opened 2022-05-03
Improvement/143 redeem popup styling

VargaElod23:


rattrap (changes requested):
Let's make the status column look like in the design and add another column for the redeem button (can be enabled or disabled)

services/community-site/src/assets/icons/greyChech.tsx
```diff
@@ -0,0 +1,10 @@
+import React from 'react';
```
rattrap, line 1: Typo in file name

services/community-site/src/assets/icons/greyChech.tsx
```diff
@@ -0,0 +1,10 @@
...
+
```
rattrap, line 2: We also have this icon already in the Sidebar, on the redeem menu.

rattrap (changes requested):

rattrap (approved):

PR #147, opened 2022-05-05
Improvement/55 redeem improvements

VargaElod23:
All design differences were eliminated, only thing that I am unsure about is the Redeem box copy text. Is the one on figma the one we wanna update to @rattrap ? 

Also, check 3: ` show all claims in the table, not just the ones that aren't claimed` not sure what's happenig there but it's working properly whenever I test: 
image

### DIFF
--- a/services/community-site/src/services/useRedeem.ts
+++ b/services/community-site/src/services/useRedeem.ts
@@ -53,11 +53,8 @@ function useRedeem() {
     transformedClaims.sort(
       (claimA, claimB) => claimA.createdAt.getTime() - claimB.createdAt.getTime(),
     );
-    const finalClaims = transformedClaims.map((_claim) => {
-      return _claim;
-    });
-    const reversedClaims = finalClaims.reverse();
-    const totalUnclaimed = finalClaims
+    transformedClaims.reverse();
+    const totalUnclaimed = transformedClaims
       .map((c) => {
         if (!c.claimed && c.numberOfTokens.gt('0')) {
           return c.numberOfTokens;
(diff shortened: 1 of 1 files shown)


rattrap (changes requested):
Looks good!

Let's rename `Redemption History` to `Redemption History` and remove the `Lifetime points redeemed` column and that's it on the UI.

Also need to modify the API to return all claims.

services/claim-backend/src/modules/claim/account.controller.ts
```diff
@@ -82,7 +83,11 @@ export class AccountController {
     @Param('account') account: string,
   ): Promise<CollectionResponse<ClaimEntity>> {
     return this.claimService.claims({
-      filter: { address: account },
+      filter: {
```
rattrap, line 86: I wouldn't add this here. It should go in the service. I don't think we need to pass this whole Query object anyway as we are hardcoding the values.

rattrap: So let's just have a simple `getClaimsFor(address: string)` method

rattrap (changes requested):

services/claim-backend/src/modules/claim/claim.service.ts
```diff
@@ -368,6 +368,19 @@ export class ClaimService {
     }));
     return claims;
   }
+  public async getClaimsFor(
+    address: string,
+  ): Promise<CollectionResponse<ClaimEntity>> {
+    const claims = new CollectionResponse<ClaimEntity>();
+    [claims.data, claims.count] = await this.claimRepository.findAndCount({
```
rattrap, line 375: We need to add the LOWER(address) thing too.

services/claim-backend/src/modules/claim/account.controller.ts
```diff
@@ -24,6 +24,7 @@ import {
 } from '@taraxa-claim/common';
 import { AccountEntity } from './entity/account.entity';
 import { ClaimEntity } from './entity/claim.entity';
+import { Raw } from 'typeorm';
```
rattrap, line 27: Not needed anymore.

rattrap (changes requested):

PR #149, opened 2022-05-09
Feature/65 new login popups

VargaElod23:


packages/taraxa-ui/src/components/ProfileCard/ProfileCard.tsx
```diff
@@ -13,13 +13,15 @@ import { toSvg } from 'jdenticon';
 
 import theme from '../theme';
 
+import ExcalamtionTriangle from '../Icons/ExclamationTriangle';
```
Malak67, line 16: Typo. `Exclamation`

services/community-site/src/assets/icons/greenCircularCheck.tsx
```diff
@@ -0,0 +1,30 @@
+import React from 'react';
+
+interface CircleCheckProps {
+  color?: string;
+}
+
+const GreenCircledCheckIcon = (props: CircleCheckProps) => {
```
Malak67, line 7: This is not exported. Shouldn`t we export both of them or separate them into two files to use the same pattern.

PR #168, opened 2022-08-18
Feature/migrate comm site to mui5

VargaElod23:


packages/taraxa-ui/src/components/DataCard/DataCard.styles.ts
```diff
@@ -9,7 +9,7 @@ const useStyles = makeStyles(() => {
       backgroundColor: '#151823',
       paddingLeft: '24px',
       paddingRight: '24px',
-      border: '1px solid #15AC5B',
+      border: '1px solid #15AC5B !important',
```
Malak67, line 12: Why do we need important here? Does it get overwritten somewhere else?

Malak67 (approved):
LGTM. Only concerned about using '!important'. we should try and avoid it as much as we can.

farebord:
> LGTM. Only concerned about using '!important'. we should try and avoid it as much as we can.

I agree with this. Sometimes is difficult with MUI

farebord (approved):
Looks good to me! Great work @VargaElod23

packages/taraxa-ui/src/components/DataCard/DataCard.styles.ts
VargaElod23: Yes, it gets overridden for some reason by the `Card` component `root` styles.

PR #178, opened 2022-08-23
Init explorer cra, adjust config

VargaElod23:


services/explorer/package.json
```diff
@@ -0,0 +1,77 @@
+{
+  "name": "@taraxa_project/explorer",
+  "description": "Taraxa Explorer",
+  "version": "0.0.1",
+  "main": "template.json",
+  "license": "MIT",
```
Malak67, line 6: I don't think this will be MIT, right?
we should keep: license": "UNLICENSED",

services/explorer/package.json
```diff
@@ -0,0 +1,77 @@
...
+  "dependencies": {
+    "@material-ui/core": "^4.12.2",
```
Malak67, line 8: Why do we have @material-ui version 4 here and also @mui/material? We should remove the old version but make sure the new version has all it's dependencies like 'lab' and 'icons'

services/explorer/src/App.css
```diff
@@ -0,0 +1,38 @@
+.App {
```
Malak67, line 1: This is left probably from CRA but we should not forget about removing it.

services/explorer/src/App.test.tsx
```diff
@@ -0,0 +1,9 @@
+import React from 'react';
```
Malak67, line 1: We won't need this test file as our App will change

Malak67 (changes requested):
Some small fixes needed.

Malak67 (approved):

PR #179, opened 2022-08-24
Feature/explorer main page

VargaElod23:


services/explorer/src/components/Header/Header.tsx
```diff
@@ -0,0 +1,65 @@
...
+
+import { HamburgerIcon, TaraxaIcon } from '../icons';
+
+import './header.scss';
+
+const Header = () => {
+  const isMobile = useMediaQuery({ query: `(max-width: 760px)` });
+  const buttons = (
+    <div className="buttonsContainer">
+      <Button label="DAG" color="primary" variant="text" onClick={() => console.log('DAG data')} />
```
Malak67, line 15: I would create a Header.effects.ts file, and create there an array of buttonsm each having the props: label, color, onAction etc. Then use that here to iterate over the array. But we can also do this later on.

services/explorer/src/App.tsx
```diff
@@ -1,25 +1,41 @@
 import React from 'react';
+import Footer from './components/Footer';
+import Header from './components/Header/Header';
 import logo from './logo.svg';
 
-function App() {
+declare global {
```
Malak67, line 6: Do we need this?

Malak67 (approved):
Left 2 comments but we can merge and handle them afterwards

PR #184, opened 2022-08-24
Fix header styles

VargaElod23:
image


packages/taraxa-ui/src/components/Header/Header.tsx
```diff
@@ -37,17 +37,19 @@ function Header({
       <AppBar {...props} variant='elevation'>
         <Container maxWidth={false}>
           <Toolbar variant='regular'>
-            {Icon && (
-              <a className={classes.headerIconContainer} href='/'>
-                <Icon />
-              </a>
-            )}
+            <div className={classes.flexContainer}>
```
Malak67, line 40: Replace this div with:
`<Box sx={{ display: 'flex', justifyContent: 'center', alignItems: 'center' }}>`

packages/taraxa-ui/src/components/Header/Header.styles.ts
```diff
@@ -3,6 +3,12 @@ import theme from '../theme';
 
 const useStyles = makeStyles(() => {
   return {
+    flexContainer: {
```
Malak67, line 6: After adding the Box component this can be removed

Malak67 (approved):

PR #187, opened 2022-08-25
Feature/implement transactions page

VargaElod23:


services/explorer/src/components/Header/Header.tsx
```diff
@@ -8,25 +8,42 @@ import {
...
 import MenuIcon from '@mui/icons-material/Menu';
+import { theme } from '../../theme-provider';
 import { TaraxaIcon } from '../icons';
 import { HeaderBtn, useHeaderEffects } from './Header.effects';
 
 export const Header = () => {
   const isMobile = useMediaQuery({ query: `(max-width: 760px)` });
-  const { headerButtons } = useHeaderEffects();
+  const { headerButtons, selected } = useHeaderEffects();
+  const buttonRow = selected
```
Malak67, line 18: This seems a bit overkill to me.
Why can't we keep the whole array of headerButtons in the state and when a location changes, we just set the state and update the current header btn as selected. Then here we don't need to iterate again and modify the property on a button.

services/explorer/.eslintrc.js
```diff
@@ -9,6 +9,20 @@ module.exports = {
       },
       rules: {
         'no-restricted-exports': 'off',
+        'prettier/prettier': [
```
Malak67, line 12: I am pretty sure this will break my code again....

services/explorer/src/pages/Transactions/Transactions.tsx
```diff
@@ -0,0 +1,26 @@
...
+import { Table } from '@taraxa_project/taraxa-ui';
+import { Container, Typography } from '@mui/material';
+import { toTableRow } from '../../components/TransactionRow/TransactionRow';
+import { useTransactionsEffects } from './Transactions.effect';
+
+const TransactionsPage = () => {
+  const { data, columns } = useTransactionsEffects();
+  const rows = data ? data.map((row) => toTableRow(row)) : [];
+  return (
+    <Container maxWidth='xl'>
```
Malak67, line 11: I think we will need this everywhere including the header and footer, right?
Can't we add it in the App.tsx as a wrapper for the whole application, which includes, header and footer?

Malak67 (changes requested):
Small changes are needed

services/explorer/src/components/TransactionRow/TransactionRow.tsx
```diff
@@ -0,0 +1,63 @@
...
+import moment from 'moment';
+import React from 'react';
+import {
+  TransactionData,
+  TransactionStatus,
+} from '../../pages/Transactions/Transactions.effect';
+
+export const toTableRow = (props: TransactionData) => {
+  const { timestamp, block, status: state, txHash, value, token } = props;
+  const txDate = moment(timestamp).format('dddd, MMMM Do, YYYY h:mm:ss A');
```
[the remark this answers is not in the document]
VargaElod23: No particular reason, to be honest

services/explorer/src/components/TransactionRow/TransactionRow.tsx
```diff
@@ -0,0 +1,63 @@
...
+  let labelType: JSX.Element;
+  if (state === TransactionStatus.SUCCESS) {
```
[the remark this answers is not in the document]
VargaElod23: No particular reason, to be honest

services/explorer/src/components/Header/Header.effects.ts
```diff
@@ -14,31 +19,57 @@ export type HeaderBtn = {
...
       label: 'Faucet',
       color: 'primary',
       variant: 'text',
-      onAction: () => console.log('Faucet data'),
+      selected: false,
+      onAction: () => onClick('faucet'),
     },
   ];
-  return { headerButtons };
+  const [selected, setSelected] = useState<HeaderBtn>(
```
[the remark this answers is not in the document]
VargaElod23: This uses the `headerButtons` data which in turn uses `onClick` so that's why they're defined in this sequence

services/explorer/.eslintrc.js
VargaElod23: It may do so, but this is a place it was missing, we should have it here too. Just reformat :D

services/explorer/src/components/Header/Header.tsx
VargaElod23: Right, I'll refactor

services/explorer/src/components/Header/Header.effects.ts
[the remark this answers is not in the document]
Malak67: Makes sense

VargaElod23:
image

Malak67 (approved):

PR #192, opened 2022-08-26
Feature/implement blocks page

VargaElod23:


VargaElod23:
image

services/explorer/src/components/TransactionRow/TransactionRow.tsx
```diff
@@ -1,13 +1,14 @@
 import { Icons, Label } from '@taraxa_project/taraxa-ui';
 import moment from 'moment';
 import React from 'react';
+import { BlockData } from '../../pages/Blocks/Blocks.effects';
```
Malak67, line 4: Ideally we should not import files from effects in components. If we use the files here, then we should have the BlockData somewhere outside, in a models folder. Same goes for the imports from Transaction.effects. Let's move these models outside (in a models folder, table.ts file) and then import them here and in the effects as well.

Malak67 (approved):
LGTM. One minor comment to move the models outside.

services/explorer/src/components/TransactionRow/TransactionRow.tsx
```diff
@@ -62,3 +63,41 @@ export const toTableRow = (props: TransactionData) => {
     ],
   };
 };
+
+export const toBlockTableRow = (props: BlockData) => {
+  const { timestamp, block, txHash, transactionCount } = props;
+  let age = Math.floor(+new Date() / 1000 - +timestamp);
```
Malak67, line 69: Can we move this in an utils function outside?

PR #205, opened 2022-09-01
Feature/implement faucet

VargaElod23:


services/faucet/src/modules/faucet/faucet.service.ts
```diff
@@ -0,0 +1,91 @@
...
+        throw new Error(
+          `You can withdraw ${remainder} TARA tokens for the current week only. Please choose an amount smaller or equal to it.`
+        );
+    }
+    const request = new RequestEntity();
+    request.address = requestDto.address;
+    request.amount = requestDto.amount;
+    request.createdAt = requestDto.timestamp;
+
+    if (request) {
```
Malak67, line 66: Not sure if you need this if statement as you created the request above

services/faucet/src/modules/faucet/faucet.service.ts
```diff
@@ -0,0 +1,91 @@
...
+      address: requestDto.address,
+    });
+
+    const requestsOfThisWeek = filterRequestsOfThisWeek(requestsForAddress);
+    if (requestsOfThisWeek && requestsOfThisWeek.length > 0) {
+      const total = requestsOfThisWeek
+        .map((r) => r.amount)
+        .reduce((curr, next) => curr + next);
+      if (total >= +this.generalConfig.maxRewardPerWeek!)
+        throw new Error(
```
Malak67, line 51: Wouldn't it be better to return more specific errors from Nestjs, or at least InternalServerErrorException?

services/faucet/src/modules/faucet/faucet.service.ts
```diff
@@ -0,0 +1,91 @@
...
+          this.ethereumConfig.contractAddress,
+          request.address,
+          ethers.utils.parseEther(`${request.amount}`)
+        );
+      if (transfer) {
+        const registered = await this.requestRepository.save(request);
+        return registered;
+      } else throw new Error(`${transfer} was unsuccessful.`);
+    }
+    throw new Error(
```
Malak67, line 87: Not sure if this is really needed here, or it can be moved at the top like if (!request) throw new Error ...

services/faucet/src/modules/faucet/faucet.service.ts
```diff
@@ -0,0 +1,91 @@
...
+        +this.generalConfig.maxRewardPerWeek! - requestDto.amount;
+      if (remainder < requestDto.amount)
+        throw new Error(
+          `You can withdraw ${remainder} TARA tokens for the current week only. Please choose an amount smaller or equal to it.`
+        );
+    }
+    const request = new RequestEntity();
+    request.address = requestDto.address;
+    request.amount = requestDto.amount;
+    request.createdAt = requestDto.timestamp;
```
Malak67, line 64: this requestDto.timestamp should be handled as specified in the frontend. We should receive a Date Object or something, convert it to the appropiate UTC time, then save it in UTC.

services/faucet/src/modules/faucet/utils/utils.ts
```diff
@@ -0,0 +1,34 @@
...
+  const maxDate = new Date();
+  maxDate.setDate(monthDay + daysToSunday);
+  setDateToMidnight(maxDate);
+  const minDate = new Date();
+  minDate.setDate(monthDay - daysFromSunday);
+  setDateToMidnight(minDate);
+
+  return requests.filter((req) => {
+    if (
+      req.createdAt.getTime() < maxDate.getTime() &&
```
Malak67, line 26: When comparing the dates here, we need to make sure that both the request date and this date are in UTC.

Malak67 (changes requested):
PR looks good. We just need to ensure that we are comparing dates in the same timezone.

services/faucet/src/modules/faucet/faucet.service.ts
VargaElod23: Fails if there is a DB connectivity issue, so it definitely helps

services/faucet/src/modules/faucet/faucet.service.ts
VargaElod23: yeah, I thought about creating some custom errors also, but this is general in all services we have for now, we just use Error. Internal could be better here, yes

services/faucet/src/modules/faucet/faucet.service.ts
Elod23: This monitors whether the tx was broadcasted successfully, so it is def needed

Malak67 (approved):

services/faucet/src/modules/faucet/faucet.service.ts
```diff
@@ -0,0 +1,103 @@
...
+        throw new InternalServerErrorException(
+          'TARA wallet cannot be initialized. Please try again later.'
+        );
+      const sendTx: ethers.providers.TransactionRequest = {
+        to: request.address,
+        value: ethers.utils.parseEther(`${request.amount}`),
+        gasPrice: 0,
+        gasLimit: 100000000,
+      };
+      const transfer = await taraWallet.sendTransaction(sendTx);
```
VargaElod23, line 80: Add queue

PR #206, opened 2022-09-02
Feature/implement faucet page

VargaElod23:


services/explorer/src/App.tsx
```diff
@@ -8,40 +8,47 @@ import BlocksPage from './pages/Blocks/Blocks';
...
-              path='/transactions/:txHash'
-              element={<TransactionDataContainer />}
-            />
-            <Route path='/nodes' element={<NodesPage />} />
-            <Route path='*' element={<Navigate to='/' replace />} />
-          </Routes>
-          <Footer />
-        </Box>
-      </Container>
+      <ExplorerLoaderProvider>
```
Malak67, line 25: I would move this in the index.tsx file, as a wrapper for the <App />. We have all the providers there so it's more consistent to move it there.

services/explorer/src/components/Footer/Footer.tsx
```diff
@@ -2,7 +2,7 @@ import React from 'react';
 import { Footer as TFooter } from '@taraxa_project/taraxa-ui';
 import { useFooterEffects } from './Footer.effects';
 
-export const Footer = () => {
+export const Footer = (): JSX.Element => {
```
Malak67, line 5: Why do we need the type here? And if we add the type, I think it should be FC imported from React, not JSX.Element.

services/explorer/src/components/Header/Header.tsx
```diff
@@ -12,7 +12,7 @@ import { theme } from '../../theme-provider';
 import { TaraxaIcon } from '../icons';
 import { HeaderBtn, useHeaderEffects } from './Header.effects';
 
-export const Header = () => {
+export const Header = (): JSX.Element => {
```
Malak67, line 15: Same here. If we add the type, let's add FC

services/explorer/src/components/Links/AddressLink.tsx
```diff
@@ -2,7 +2,7 @@ import { theme } from '@taraxa_project/taraxa-ui';
 import React from 'react';
 import { Link } from 'react-router-dom';
 
-export const AddressLink = ({ address }: { address: string }) => {
+export const AddressLink = ({ address }: { address: string }): JSX.Element => {
```
Malak67, line 5: Same, change it to FC

services/explorer/src/components/Links/TransactionLink.tsx
```diff
@@ -8,7 +8,7 @@ export const TransactionLink = ({
 }: {
   txHash: string;
   wrap?: boolean;
-}) => {
+}): JSX.Element => {
```
Malak67, line 11: Same here, FC

services/explorer/src/index.tsx
```diff
@@ -1,6 +1,7 @@
 import React from 'react';
 import ReactDOM from 'react-dom/client';
 import { BrowserRouter } from 'react-router-dom';
+import { SnackbarProvider } from 'notistack';
```
Malak67, line 4: I didn't see a snackbar in the design. If we introduce this, we should verify with the design team and ask for some design examples.
I would also recommend this: react-toastify if we need more customization but maybe notistack can work as well. Just make sure this is agreed upon.

services/explorer/src/pages/BlockData/BlockDataContainer.tsx
```diff
@@ -12,7 +12,7 @@ import {
 import { useBlockDataContainerEffects } from './BlockDataContainer.effects';
 import { BlockTable } from './BlockTable';
 
-const BlockDataContainer = () => {
+const BlockDataContainer = (): JSX.Element => {
```
Malak67, line 15: FC

services/explorer/src/pages/Blocks/Blocks.tsx
```diff
@@ -4,7 +4,7 @@ import { toBlockTableRow } from '../../utils/TransactionRow';
 import { useBlockEffects } from './Blocks.effects';
 import { PageTitle } from '../../components';
 
-const BlocksPage = () => {
+const BlocksPage = (): JSX.Element => {
```
Malak67, line 7: FC

services/explorer/src/pages/Faucet/Faucet.effects.ts
```diff
@@ -0,0 +1,105 @@
+import { useSnackbar } from 'notistack';
+import { useCallback, useEffect, useState } from 'react';
+import { faucetUri } from '../../config';
+import { useExplorerNetwork } from '../../hooks/useExplorerNetwork';
+import { useExplorerLoader } from '../../hooks/useLoader';
+
+export enum RequestLimit {
```
Malak67, line 7: I would move all enums inside utils/enums or something similar.

services/explorer/src/pages/Faucet/Faucet.effects.ts
```diff
@@ -0,0 +1,105 @@
...
+  const { enqueueSnackbar, closeSnackbar } = useSnackbar();
+  const [address, setAddress] = useState<string>('');
+  const [amount, setAmount] = useState<RequestLimit>(1);
+  const [displayToast, setDisplayToast] = useState<{
+    display: boolean;
+    variant?: 'success' | 'error' | 'warning' | undefined;
+    text?: string;
+  }>({ display: false, variant: undefined, text: '' });
+
+  const sendTokenRequest = useCallback(() => {
```
Malak67, line 25: It might be a good idea to move this outside, in a api folder. All API requests should be there.

services/explorer/src/pages/Faucet/Faucet.effects.ts
```diff
@@ -0,0 +1,105 @@
...
+      try {
+        const response = await fetch(faucetUri, {
+          method: 'POST',
+          headers: {
+            'Content-Type': 'application/json',
+          },
+          body: JSON.stringify({
+            address,
+            amount,
+            timestamp: Date.now() / 1000,
```
Malak67, line 37: We should ensure that address and amount are there so we don't do a request with empty or wrong values.
I think it might be good to add react-hook-form or something and yup for validating the address format.

services/explorer/src/pages/Faucet/Faucet.effects.ts
```diff
@@ -0,0 +1,105 @@
...
+  ONE = 1,
+  TWO = 2,
+  FIVE = 5,
+  SEVEN = 7,
+}
+export const useFaucetEffects = () => {
+  const { currentNetwork } = useExplorerNetwork();
+  const { isLoading, initLoading, finishLoading } = useExplorerLoader();
+  const { enqueueSnackbar, closeSnackbar } = useSnackbar();
+  const [address, setAddress] = useState<string>('');
```
Malak67, line 17: These should not be addres inside the state, because the only way to know when these are added inside the state is to use, useEffect which will not work as expected here. The address and amount should be handled by a form and on submit should call the request method from the API folder.

services/explorer/src/pages/Faucet/Faucet.effects.ts
```diff
@@ -0,0 +1,105 @@
...
+          headers: {
+            'Content-Type': 'application/json',
+          },
+          body: JSON.stringify({
+            address,
+            amount,
+            timestamp: Date.now() / 1000,
+          }),
+        });
+        if (!response)
```
Malak67, line 40: There are quite a few branches introduced here with these if/else checks. Should we show the error message or success message from the API? And if that is missing we show a default error notification?

services/explorer/src/pages/Faucet/Faucet.tsx
```diff
@@ -0,0 +1,102 @@
...
+      <Box
+        display='flex'
+        flexDirection='row'
+        width='100%'
+        justifyContent='center'
+        alignItems='flex-start'
+        marginBottom='6rem'
+        gap='2rem'
+      >
+        <FormControl>
```
Malak67, line 53: Here, we need a form.

services/explorer/src/pages/Home/ChartContainer.tsx
```diff
@@ -4,7 +4,7 @@ import { BarChart } from '@taraxa_project/taraxa-ui';
 import { theme } from '../../theme-provider';
 import { useHomeEffects } from './Home.effects';
 
-const ChartContainer = () => {
+const ChartContainer = (): JSX.Element => {
```
Malak67, line 7: FC

services/explorer/src/pages/Home/Home.tsx
```diff
@@ -6,7 +6,7 @@ import { PageTitle } from '../../components';
 import { useHomeEffects } from './Home.effects';
 import ChartContainer from './ChartContainer';
 
-const HomePage = () => {
+const HomePage = (): JSX.Element => {
```
Malak67, line 9: FC

services/explorer/src/pages/Nodes/Nodes.tsx
```diff
@@ -4,7 +4,7 @@ import { Box } from '@mui/material';
 import { PageTitle } from '../../components';
 import { useNodesEffects } from './Nodes.effects';
 
-const NodesPage = () => {
+const NodesPage = (): JSX.Element => {
```
Malak67, line 7: FC

services/explorer/src/pages/TransactionData/TransactionData.tsx
```diff
@@ -12,7 +12,7 @@ import { statusToLabel, timestampToAge } from '../../utils/TransactionRow';
 import { useTransactionDataContainerEffects } from './TransactionData.effects';
 import { DagTable } from './DagTable';
 
-const TransactionDataContainer = () => {
+const TransactionDataContainer = (): JSX.Element => {
```
Malak67, line 15: FC

services/explorer/src/pages/Transactions/Transactions.tsx
```diff
@@ -4,7 +4,7 @@ import { toTransactionTableRow } from '../../utils/TransactionRow';
 import { useTransactionEffects } from './Transactions.effects';
 import { PageTitle } from '../../components';
 
-const TransactionsPage = () => {
+const TransactionsPage = (): JSX.Element => {
```
Malak67, line 7: FC

services/explorer/src/utils/TransactionRow.tsx
```diff
@@ -8,7 +8,7 @@ import {
   TransactionStatus,
 } from '../models/TableData';
 
-export const statusToLabel = (state: TransactionStatus) => {
+export const statusToLabel = (state: TransactionStatus): JSX.Element => {
```
Malak67, line 11: FC

services/explorer/src/pages/Faucet/Faucet.effects.ts
Malak67, line 37: We need to ensure that the timezone sent from the frontend is the same as the one in the backend. If the user is in GMT+2 but the server is in the UTC, we need to know this when comparing the dates. This can be done directly on the backend, but I thought it is worth mentioning it here so we don't forget.
Anyways, by doing Date.now() and dividing it by 1000 you might loose the timezone. Maybe we can send directly the date (new Date()) and handle the conversion on the backend.

Malak67 (changes requested):
Code looks good but there are a few changes requested. Let me know what you think!

services/explorer/src/components/Footer/Footer.tsx
VargaElod23: FC does need much more than what we have in most of these propless components: `Type 'Element' provides no match for the signature '(props: {}, context?: any): ReactElement<any, any>'.`

services/explorer/src/components/Header/Header.tsx
VargaElod23: Same

services/explorer/src/components/Links/AddressLink.tsx
VargaElod23: Same

services/explorer/src/components/Links/TransactionLink.tsx
VargaElod23: Same

services/explorer/src/index.tsx
VargaElod23: I showcased this during Friday's design meeting and everyone was fine with it :)

services/explorer/src/components/Footer/Footer.tsx
Malak67: Got it. Ok then but if we have props then we need to switch to FC<PropInterface>

services/explorer/src/index.tsx
Malak67: Perfect

Malak67 (approved):

PR #207, opened 2022-09-02
Address Info component

farebord:
This PR adds AddressInfo and BlockTable as new components of the taraxa UI package.


packages/taraxa-ui/src/components/BlockTable/BlockTable.tsx
```diff
@@ -0,0 +1,250 @@
...
+  gas: string;
+  gasPrice: string;
+  to: string;
+  nonce: number;
+  transactionLink: JSX.Element;
+  addressLinkFrom: JSX.Element;
+  addressLinkTo: JSX.Element;
+}
+
+export const statusToLabel = (state: TransactionStatus) => {
```
VargaElod23, line 50: If this is getting extracted into a component this needs to go into utils

VargaElod23 (approved):
LGTM. A suggestion added and I have a pretty hard time figuring out why we should extract these two data containers here but they don't hurt, I guess...

packages/taraxa-ui/src/components/BlockTable/BlockTable.tsx
```diff
@@ -0,0 +1,250 @@
...
+      />
+    );
+  }
+  if (state === TransactionStatus.FAILURE) {
+    return (
+      <Label variant='error' label='Failure' gap icon={<RedCircledCancel />} />
+    );
+  }
+  return (
+    <Label variant='error' label='Failure' gap icon={<RedCircledCancel />} />
```
Malak67, line 67: This creates a problem in the UI. While the data is loading, this shows as Error and then changes to it's correct status. We should have a Loading label, with a grey color showing a small spinner. This needs to be fixed here and also in the explorer where this is used.

packages/taraxa-ui/src/components/BlockTable/BlockTable.tsx
Malak67: @farebord I tend to agree with @VargaElod23 . This is more than just a component. It's basically a whole container. Moving this inside the taraxa-ui might be too much. From what I saw in the design, we use this type of component in 3 different pages, like transactions/xxx, block/xxx, address/xxx. We should extract the small components used in all 3 pages, if there are any, and keep this container in the explorer. Because if we need another page which is similar 90% but has a small change, we either need to extend this in taraxa-ui and make sure it doesn't brake in other pages, or create a new component, both would be too much. What do you think?

Malak67: Or, alternatively, make it really reusable with custom table columns and rows. Might be a bit more timeconsuming and not sure how customizable it needs to be.

farebord: I'll try to make this more customizable and move utils to the explorer. The point is there are no shared utils in taraxa-UI and it seems they should be there already. @Malak67 @VargaElod23

Malak67: Yes, I also looked for it. You can leave it there or create an utils and export it so we can use them in explorer as well.

services/explorer/src/pages/AddressInfo/AddressInfo.tsx
```diff
@@ -0,0 +1,32 @@
+import React from 'react';
+import { Box } from '@mui/material';
+import { AddressInfo } from '@taraxa_project/taraxa-ui';
+import { PageTitle } from '../../components';
+import { useAddressInfoEffects } from './AddressInfo.effects';
+
+const AddressInfoPage = () => {
```
Malak67, line 7: Can you reuse the AddressInfo component and BlockTable for the other pages as well (Transactions and Blocks)?. It probabbly needs some customization but I think it will really help.

VargaElod23: Yep, if we extract it to the Taraxa UI we need to rename it so that the naming corresponds with its purpose: reuse to display `detailed data about transactions, blocks and addresses` : dunno what to call it, but it will definitely need some prop changes. As I foresee, besides these 3 categories there will be another data container for smart contracts, just like on Etherscan, but maybe you could refactor it to have 3 variants: transaction, block or address, and then handle the display discrepancies internally.

VargaElod23 (changes requested):

services/explorer/src/components/BlockTable/BlockTable.tsx
```diff
@@ -0,0 +1,242 @@
...
+  };
+
+  const handleChangeRowsPerPage = (
+    event: React.ChangeEvent<HTMLInputElement>
+  ) => {
+    setRowsPerPage(parseInt(event.target.value, 10));
+    setPage(0);
+  };
+  return (
+    <ThemeProvider theme={theme}>
```
Malak67, line 85: This needs to be removed from here (probably left from taraxa-ui)

services/explorer/src/components/BlockTable/BlockTable.tsx
```diff
@@ -0,0 +1,242 @@
...
+      <CssBaseline />
```
Malak67, line 86: same for this one

services/explorer/src/pages/AddressInfo/AddressInfo.effects.tsx
```diff
@@ -0,0 +1,30 @@
...
+  dagLevel: 'string',
+  dagHash: 'string',
+  value: '0.000000',
+  from: '0xc26f6b31a5...',
+  gasLimit: 'string',
+  gas: '0.000021',
+  gasPrice: 'string',
+  to: '0xc3r17b31a5...',
+  nonce: 153,
+  transactionLink: <span>0x00e193a154...</span>,
```
Malak67, line 17: This is wrong. We will not be getting the results like this from Graphql.
Please check TransactionData.effects.ts, inside useEffect and create the objects in a similar way.

services/explorer/src/pages/AddressInfo/AddressInfo.effects.tsx
```diff
@@ -0,0 +1,30 @@
+import React, { useEffect, useState } from 'react';
+
+const blockData: any[] = Array(20).fill({
```
Malak67, line 3: We don't want any here. There should be a type.

services/explorer/src/components/AddressInfo/AddressInfo.tsx
```diff
@@ -0,0 +1,129 @@
...
+  dagBlocks,
+  pbftBlocks,
+}: AddressInfoProps) => {
+  const classes = useStyles();
+  const addressIcon = toSvg(address, 40, { backColor: '#fff' });
+
+  const pricePerTara = parseFloat(balance) / parseFloat(value);
+
+  return (
+    <ThemeProvider theme={theme}>
```
Malak67, line 41: This needs to be removed. We have the theme globally

services/explorer/src/components/AddressInfo/AddressInfo.tsx
```diff
@@ -0,0 +1,129 @@
...
+      <CssBaseline />
```
Malak67, line 42: Same here. Needs to be removed

services/explorer/src/components/BlockTable/BlockTable.tsx
```diff
@@ -0,0 +1,242 @@
...
+import { theme } from '../../theme-provider';
+import { AddressLink } from '../Links';
+
+export interface TransactionDataItem extends TransactionData {
+  txHash: string;
+  dagLevel: string;
+  dagHash: string;
+}
+
+export const statusToLabel = (state: TransactionStatus) => {
```
Malak67, line 26: I don't think you need this hwere. You can reuse it from utils/TransactionRow.ts

services/explorer/src/components/BlockTable/BlockTable.tsx
```diff
@@ -0,0 +1,242 @@
...
+};
+
+export interface BlockTableProps {
+  blockData: TransactionDataItem[];
+  onFilter?: () => void;
+  onDAGFilter?: () => void;
+  onPBFTFilter?: () => void;
+}
+
+export const BlockTable: React.FC<BlockTableProps> = (props) => {
```
Malak67, line 64: Would have been prettier to extract the props directly here, and without giving them a default value.
`export const BlockTable: React.FC<BlockTableProps> = ({blockData , onFilter, onDAGFilter , onPBFTFilter }: BlockTableProps =>`

Malak67 (changes requested):
Left some comments

Malak67:
@farebord After applying the fixes requested, please fix the conflicts with the design-v2 branch.

services/explorer/src/pages/AddressInfo/AddressInfo.effects.tsx
farebord: Ohh I forgot about this one. It was just a mapper object.

Malak67 (approved):

PR #218, opened 2022-09-13
Feat/side nav

Malak67:
Extended ui table component
Created sidenav
image

### DIFF
--- a/packages/taraxa-ui/src/components/Header/Header.styles.ts
+++ b/packages/taraxa-ui/src/components/Header/Header.styles.ts
@@ -4,7 +4,7 @@ import theme from '../theme';
 const useStyles = makeStyles(() => {
   return {
     headerIconContainer: {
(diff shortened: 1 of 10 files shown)


packages/taraxa-ui/src/components/Header/Header.styles.ts
```diff
@@ -4,7 +4,7 @@ import theme from '../theme';
 const useStyles = makeStyles(() => {
   return {
     headerIconContainer: {
-      backgroundColor: '#202534',
+      backgroundColor: theme.palette.grey[800],
       border: '1px solid #40465F',
```
farebord, line 8: We can use the palette for the border color as well.

farebord (changes requested):
Small change request.

farebord (approved):
LGTM

PR #224, opened 2022-09-16
Home page charts

farebord:
This PR removes mocks and uses real data to render the graphs.
Fix mock types.


services/explorer/src/pages/Home/Home.tsx
```diff
@@ -17,7 +17,7 @@ const HomePage = () => {
         title={currentNetwork}
         subtitle='Search for addresses, block hashes, transactions...'
       />
-      <ChartContainer />
+      <ChartContainer pbftBlocks={pbftBlocks} dagBlocks={dagBlocks} />
```
Malak67, line 20: Are we sure that for the charts we need the same pbft and dag blocks as we do for the cards? For the cards I am only using the last 10 blocks. But for charts, I don't know if we need just the last 10 or more. Can you please double-check this before merging?

services/explorer/src/pages/Home/ChartContainer.tsx
```diff
@@ -2,14 +2,23 @@ import React from 'react';
 import { Grid } from '@mui/material';
 import { BarChart } from '@taraxa_project/taraxa-ui';
 import { theme } from '../../theme-provider';
-import {
-  getMockedDagBlocksCard,
-  getMockedPbftBlocksCard,
-} from '../../api/mocks';
+import { DagBlock, PbftBlock } from '../../models';
+// import {
```
Malak67, line 6: If we don't need these anymore, you can delete them. If you might need them to test something, then leave them commented.

services/explorer/src/pages/Home/ChartContainer.tsx
```diff
@@ -2,14 +2,23 @@ import React from 'react';
...
+export interface ChartContainerProps {
+  dagBlocks: DagBlock[];
+  pbftBlocks: PbftBlock[];
+}
+
+const ChartContainer = ({
+  dagBlocks,
+  pbftBlocks,
+}: ChartContainerProps): JSX.Element => {
+  // const dagBlocks = getMockedDagBlocksCard();
```
Malak67, line 20: Same here

services/explorer/src/pages/Home/ChartContainer.tsx
```diff
@@ -26,18 +35,28 @@ const ChartContainer = (): JSX.Element => {
...
+          labels={pbftBlocks.slice(5).map((block) => block.number.toString())}
           datasets={[
             {
-              data: dagBlocks
-                .concat(pbftBlocks)
-                .sort((a, b) => +b.timestamp - +a.timestamp)
-                .slice(0, 6)
-                .flatMap((x) => +x.transactionCount),
+              data: pbftBlocks.slice(5).map((block) => {
+                console.log(
```
Malak67, line 42: Do we still need the console.log or we can remove it?

services/explorer/src/pages/Home/Home.tsx
VargaElod23: Yo guys I'm back. On the charts each vertical bar should represent a block, therefore you can visually count how much we display or even make it dynamic so it displays as much as it's given.

services/explorer/src/pages/Home/ChartContainer.tsx
```diff
@@ -46,43 +65,43 @@ const ChartContainer = (): JSX.Element => {
...
+                    .timestamp
+                );
+              }),
               borderRadius: 5,
               barThickness: 20,
               backgroundColor: theme.palette.secondary.main,
             },
           ]}
         />
-        <BarChart
```
VargaElod23, line 67: 🧹

services/explorer/tsconfig.json
```diff
@@ -1,46 +1,46 @@
...
-  "include": [
-    "**/src/**/*"
-  ],
-  "exclude": [
-    "node_modules",
-    "build",
-    "*.d.ts"
-  ]
-}
+{
```
VargaElod23, line 1: @farebord please ensure that you use CRLF so that formatting changes like these won't occur anymore + EOF newline

VargaElod23 (approved):
LGTM, small cleanup needed 👍🏼

services/explorer/src/pages/Home/Home.tsx
farebord: @Malak67 @VargaElod23 I'm just passing pbftBlocks and dagBlocks entirely to the ChartContainer. We probably could be okay just passing them 6 elements of each array since we need the previous block for each block we are iterating to calculate block time.

services/explorer/src/pages/Home/ChartContainer.tsx
```diff
@@ -81,44 +71,55 @@ const ChartContainer = ({
...
           ]}
         />
-        {/* <BarChart
+        <BarChart
           title='DAG efficiency'
           labels={dagBlocks
-            .sort((a, b) => +b.timestamp - +a.timestamp)
-            .slice(0, 6)
-            .flatMap((x) => x.level)}
+            .slice(6)
```
Malak67, line 95: I think we can leave the last 10 instead of 6 to be consistent with the two cards. @VargaElod23  what do you think?

services/explorer/src/pages/Home/ChartContainer.tsx
```diff
@@ -81,44 +71,55 @@ const ChartContainer = ({
...
             {
-              data: dagBlocks?.slice(5).map((block) => {
-                return (
-                  block.timestamp -
-                  dagBlocks.find((b) => b.pbftPeriod === block.pbftPeriod - 1)
-                    .timestamp
-                );
-              }),
+              data: dagBlocks
+                ?.slice(0, 5)
```
Malak67, line 76: Same here. I would keep the last 10.

services/explorer/src/pages/Home/ChartContainer.tsx
VargaElod23: I agree. or even make it dynamic and set it to 10 by default

VargaElod23:
image

packages/taraxa-ui/src/components/BarChart/BarChart.tsx
```diff
@@ -23,7 +23,7 @@ ChartJS.register(
   Legend
 );
 
-const DEFAULT_OPTIONS = {
+const useTick = (tick: string) => ({
```
Malak67, line 26: I would rename this because if you use the word 'use' it is considered as a React Hook. getTick or returnTick or something similar.

services/explorer/src/pages/Home/ChartContainer.tsx
```diff
@@ -46,36 +54,41 @@ const ChartContainer = (): JSX.Element => {
...
+            .slice(0, 5)
+            .map((block) => block.number.toString())}
           datasets={[
             {
-              data: dagBlocks
-                .concat(pbftBlocks)
-                .sort((a, b) => +b.timestamp - +a.timestamp)
-                .slice(0, 6)
-                .flatMap((x) => +x.transactionCount),
+              data: calculatePBFTBlockTime(pbftBlocks.slice(0, 6)),
```
Malak67, line 63: Shouldn`t this take 6 elements? So it should be slice(0,5)

Malak67 (approved):
Looks good! Just some minor comments.

PR #239, opened 2022-09-20
Feature/232 route block numbers to details page

VargaElod23:
image
image
Hey guys! A first minimalist version is ready for both pages. :) 👍🏼


services/explorer/src/api/graphql/queries/dagBlocks.ts
```diff
@@ -1,15 +1,54 @@
...
-    dagBlock (hash: $hash) {
-      hash,
-      pivot,
-      tips,
-      level,
-      pbftPeriod,
-      timestamp,
-      transactionCount,
+query dag_block_query($hash: Bytes32) {
+  dagBlock (hash: "0c9d36dd14c50e0c319e095f06698a05bb80def51c7927d783514b051c619454") {
```
Malak67, line 3: You probabbly forgot this when testing.
We need to have the variable $hash here

services/explorer/src/api/graphql/queries/pbftBlocks.ts
```diff
@@ -7,8 +7,24 @@ export const blockQuery = `
       gasLimit,
       gasUsed,
       timestamp,
+			parent {
```
Malak67, line 10: The spacing seems a bit weird. Is it like this in the code or just in the PR?

services/explorer/src/models/TableData.ts
```diff
@@ -1,6 +1,6 @@
 export enum TransactionStatus {
-  SUCCESS = '0x0',
-  FAILURE = '0x1',
+  SUCCESS = '0x1',
```
Malak67, line 2: This is not really used to be honest. I saw that the status is either 0, 1 or null. I created a new method to convert this (formatTransactionStatus).
Just double check if with this change, the status show the correct status (I think it was a component in taraxa-ui)

services/explorer/src/pages/BlockData/DAGDataContainer.effects.tsx
```diff
@@ -0,0 +1,45 @@
...
+    query: dagDetailsQuery,
+    variables: {
+      hash,
+    },
+    pause: !hash,
+  });
+  const { initLoading, finishLoading } = useExplorerLoader();
+
+  const fetchBlockDetails = useCallback(() => {
+    setTimeout(() => {
```
Malak67, line 24: You don't need the setTimeout anymore here. And this can be wrapped inside an useEffect like in the other pages.

services/explorer/src/pages/BlockData/DAGDataContainer.effects.tsx
```diff
@@ -0,0 +1,45 @@
...
+      console.log('dddd: ', data);
```
Malak67, line 25: We should have the console.log removed

services/explorer/src/pages/BlockData/DAGDataContainer.effects.tsx
```diff
@@ -0,0 +1,45 @@
...
+      setBlockData(data?.dagBlock);
```
Malak67, line 26: This should be inside an if statement if there is data?.dagBlock - sometimes it comes as null

services/explorer/src/pages/BlockData/DAGDataContainer.effects.tsx
```diff
@@ -0,0 +1,45 @@
...
+      setTransactions(data?.dagBlock?.transactions);
+    }, 500);
+  }, [data]);
+
+  useEffect(() => {
+    if (fetching) {
+      initLoading();
+    } else {
+      fetchBlockDetails();
```
Malak67, line 35: You don't need to fetch them here. If the data is inside a useEffect, it is enough. Here you just need to show and hide the loading spinner.

services/explorer/src/pages/BlockData/DAGDataContainer.tsx
```diff
@@ -98,21 +98,23 @@ const BlockDataContainer = (): JSX.Element => {
               <HashLink
                 linkType={HashLinkType.ADDRESSES}
                 width='auto'
-                hash={blockData?.sender}
+                hash={zeroX(blockData?.author?.address)}
```
Malak67, line 101: Are you sure we need to add '0x' in front of the address?

services/explorer/src/pages/PBFTData/PBFTDataContainer.effects.tsx
```diff
@@ -1,40 +1,45 @@
...
-    }, 1000);
-  }, []);
+    }
+  }, [currentNetwork, fetching]);
 
   useEffect(() => {
-    fetchBlockDetails();
-    fetchTransactions();
-  }, [currentNetwork]);
+    if (data && data?.block) {
```
Malak67, line 37: you don't need to check for both if you use data?.block.  So it is enough to check if(data?.block)

services/explorer/src/pages/PBFTData/PBFTDataContainer.effects.tsx
```diff
@@ -1,40 +1,45 @@
...
+      console.log(data);
```
Malak67, line 38: Don't forget about the console.log

services/explorer/src/pages/PBFTData/PBFTDataContainer.tsx
```diff
@@ -70,12 +70,12 @@ const PBFTDataContainer = (): JSX.Element => {
               component='h6'
               style={{ fontWeight: 'bold', wordBreak: 'break-all' }}
             >
-              {identifier}
+              {zeroX(blockData.hash)}
```
Malak67, line 73: Double-check if we need to add the 0x in front of the hash. I don't think we do.
Also, we need to make sure the blockData is here otherwise it will crash. So add blockData?.hash. And do this in all DataRow elements where it is possible.
I had to do the same thing in TransactionData.tsx because it crashed.

services/explorer/src/utils/TransactionRow.tsx
```diff
@@ -6,6 +6,13 @@ import { HashLink } from '../components/Links';
 import { BlockData, TransactionTableData, TransactionStatus } from '../models';
 import { HashLinkType } from './Enums';
 
+export const toHexStatus = (status: number): TransactionStatus =>
```
Malak67, line 9: You don't need this. We already have the formatTransactionStatus method in utils. If needed, update that method and remove this one. But make sure it works the same where it is used.

Malak67 (changes requested):
Nice job!
Some changes are required.

services/explorer/src/api/graphql/queries/pbftBlocks.ts
VargaElod23: Strange, it seems its just in the PR

services/explorer/src/models/TableData.ts
VargaElod23: Yeah, it was switched-up beforehand. Should be fine now.

services/explorer/src/pages/BlockData/DAGDataContainer.tsx
VargaElod23: I was adding it just for displaying, but you're right, I'm moving it into the displaying component

services/explorer/src/pages/BlockData/DAGDataContainer.effects.tsx
```diff
@@ -0,0 +1,42 @@
...
+    query: dagDetailsQuery,
+    variables: {
+      hash,
+    },
+    pause: !hash,
+  });
+  const { initLoading, finishLoading } = useExplorerLoader();
+
+  const fetchBlockDetails = useCallback(() => {
+    setBlockData(data?.dagBlock);
```
Malak67, line 24: I would still add an if statement here to ensure data is not null.

services/explorer/src/pages/PBFTData/PBFTDataContainer.effects.tsx
```diff
@@ -1,40 +1,44 @@
...
+    if (fetching) {
+      initLoading();
+    } else {
       finishLoading();
-    }, 1000);
-  }, []);
+    }
+  }, [currentNetwork, fetching]);
 
   useEffect(() => {
```
Malak67, line 36: I would change in DagEffects the useCallback and do it like this, or change it here to be consistent

Malak67 (approved):
Love it!

PR #242, opened 2022-09-22
Feature/241 implement address graphql fetch

VargaElod23:


services/explorer/src/pages/AddressInfo/AddressInfo.effects.tsx
```diff
@@ -1,35 +1,68 @@
...
+export const useAddressInfoEffects = (account: string) => {
+  const { currentNetwork } = useExplorerNetwork();
   const [transactions, setTransactions] = useState<Transaction[]>();
   const [dagBlocks, setDagBlocks] = useState<BlockData[]>();
   const [pbftBlocks, setPbftBlocks] = useState<BlockData[]>();
   const [addressInfoDetails, setAddressInfoDetails] =
     useState<AddressInfoDetails>();
+  const [{ fetching, data }] = useQuery({
+    query: accountQuery,
+    variables: cleanDeep({
```
Malak67, line 24: Not sure if you need cleanDeep here. Because account is always needed, right? And the hook won't work if account is missing (because of pause). So I don't think you need cleanDeep here

services/explorer/src/pages/AddressInfo/AddressInfo.effects.tsx
```diff
@@ -1,35 +1,68 @@
...
-      setAddressInfoDetails(addressInfoProps);
-    }, 500);
-  }, []);
+  useEffect(() => {
+    if (fetching) {
+      initLoading();
+    } else {
+      finishLoading();
+    }
+  }, [fetching, currentNetwork]);
```
Malak67, line 37: You can remove currentNetwork from here. If we change networks this might get triggered and we don't want that. Just use fetching here.

Malak67 (approved):
LGTM! Just 2 minor changes to fix

PR #246, opened 2022-09-26
Feat/community fixes

Malak67:
image

The main visible change is the card buttons and sizes in the Staking page (left is old version)


VargaElod23:
Tested & the cards still look good. LGTM 👍🏼

PR #257, opened 2022-10-03
feat(taraxa-models): Created models package

Malak67:


VargaElod23 (approved):
Lg tho we won't need that start command

PR #258, opened 2022-10-04
Feature/add mighrations for faucet

VargaElod23:
image


services/faucet/src/general.module.ts
```diff
@@ -27,7 +42,9 @@ import { APP_GUARD } from '@nestjs/core';
         password: configService.get<string>('database.pass'),
         database: configService.get<string>('database.name'),
         entities: [`${__dirname}/**/*.entity{.ts,.js}`],
-        synchronize: true,
+        synchronize: false,
+        autoLoadEntities: true,
+        logging: ['info'],
```
Malak67, line 47: We might need the ssl part here.
But, what I noticed is that if you add it here, it will work when deploying it but it won't work anymore locally. Hence, this condition should be added conditionally. Ex:
```
 if (!!process.env.DATABASE_CERT) {
    typeOrmOptions = {
      ...baseConnectionOptions,
      ssl: {
        rejectUnauthorized: false,
        ca: process.env.DATABASE_CERT,
      },
    };
  } else {
    typeOrmOptions = { ...baseConnectionOptions };
  }
```

Malak67 (approved):
Noice!!!

PR #263, opened 2022-10-05
Explorer: Finalize Alpha 

VargaElod23:


services/explorer-indexer/.env.example
```diff
@@ -1,9 +1,12 @@
 # --- Port used by backend to run the server --- #
 SERVER_PORT=3040
-
+WS_SERVER_PORT=80
 # --- DATABASE --- #
 DB_HOST=localhost
-DB_PORT=5432
-DB_USERNAME=postgres
+DB_PORT=55000
```
Malak67, line 6: Shouldn't this be 5432? That`s the default port of postgres. If locally it is custom, we use than in the .env but in the .env.example we should keep the default one.

services/explorer-indexer/src/modules/node/nodeSyncer.service.ts
```diff
@@ -0,0 +1,215 @@
...
+          Topics.NEW_PBFT_BLOCKS,
+          //   Topics.NEW_HEADS,
+        ],
+      }),
+      (err: Error) => {
+        if (err) this.logger.error(err);
+      }
+    );
+    this.logger.warn(
+      `Subscribet to eth_subscription method ${Topics.NEW_PBFT_BLOCKS}`
```
Malak67, line 106: Subscribed

services/explorer-indexer/src/modules/node/nodeSyncer.service.ts
```diff
@@ -0,0 +1,215 @@
...
+  @EventListener('close')
+  onClose(code: number) {
+    this.logger.error(
+      `Server at ${this.ws.url} disconnected with code ${code}`
+    );
+  }
+
+  @EventListener('ping')
+  onPing(data: Buffer) {
+    const j = data.toJSON();
```
Malak67, line 139: Let's use descriptive variable names. Or at leased 'parsedData'

services/explorer-indexer/src/modules/pbft/pbft.entity.ts
```diff
@@ -18,7 +18,7 @@ export class PbftEntity extends BaseEntity implements IPBFT {
   @IsNotEmpty()
   number: number;
 
-  @Column({ nullable: false })
+  @Column({ nullable: false, default: 0 })
```
Malak67, line 21: Don't see why we should have default 0 as timestamp. Will there be a case when we will have a pbft block without timestamp and we want to have it 0?

services/explorer-indexer/src/modules/node/nodeSyncer.service.ts
```diff
@@ -0,0 +1,215 @@
...
+import {
+  checkType,
+  NewDagBlockFinalizedResponse,
+  NewDagBlockResponse,
+  NewPbftBlockResponse,
+  ResponseTypes,
+  toObject,
+} from 'src/types';
+
+export enum Topics {
```
Malak67, line 20: I would move these two enums to the types file as well.

services/explorer-indexer/src/modules/pbft/pbft.entity.ts
VargaElod23: Yes, there are data packets where it comes as `undefined`.

services/explorer-indexer/src/modules/dag/dag.service.ts
```diff
@@ -59,4 +65,15 @@ export default class DagService {
     const updated = await this.dagRepository.save(dag);
     if (updated) this.logger.log(`DAG ${updateData.block} finalized`);
   }
+
+  private async updateNode(dag: DagEntity) {
+    const foundNode = await this.nodeRepository.findOne({
+      where: { address: dag.author },
+    });
+    if (foundNode) {
+      foundNode.dagCount += 1;
```
VargaElod23, line 74: It might be a more robust way of doing this to implement a query that returns the # of dags where the node's address is its author. WDYT @Malak67 ?

services/explorer-api/src/modules/pbft/pbft.service.ts
```diff
@@ -0,0 +1,32 @@
...
+      .startOf('week')
+      .toSeconds()
+      ?.toString()
+      .split('.')[0];
+    const sunday = DateTime.now()
+      .endOf('week')
+      .toSeconds()
+      ?.toString()
+      .split('.')[0];
+    const total = await this.repository.findBy({
```
VargaElod23, line 27: 👍🏼

services/explorer/src/api/fetchNumberBlocks.explorer-api.ts
```diff
@@ -0,0 +1,28 @@
...
+import { API } from './types';
+
+const getBlocksNumber = () => {
+  const url = `${API}/pbft/total-this-week`;
+  return axios.get(url);
+};
+
+export const useGetBlocksThisWeek = () => {
+  const { data, isError, error, isLoading, isFetching } = useQuery(
+    ['pbgt-blocks-this-week'],
```
VargaElod23: `['pbft-blocks-this-week']`

services/explorer-api/src/modules/node/node.service.ts
```diff
@@ -30,6 +31,16 @@ export class NodeService {
     };
   }
 
+  public async findByAddress(address: string): Promise<TaraxaNode> {
+    const node = await this.repository.findOneBy({ address: address });
```
VargaElod23, line 35: `findOneOrFail`  instead of if ?

Malak67: Works but I prefer to handle and throw my own expections in cases like this. That`s why I used this

VargaElod23: OK. W/e you prefer

services/explorer-indexer/src/modules/dag/dag.service.ts
Malak67: Yeah, I am also not that happy with this way of doing it. We can discuss it and see what other options we have.

Malak67 (approved):

PR #266, opened 2022-10-10
Feature/historical sync

VargaElod23:


services/explorer-indexer/src/modules/dag/dag.service.ts
```diff
@@ -41,35 +43,67 @@ export default class DagService {
...
+    const txes = await findTransactionsByHashesOrFill(
       transactions,
-    };
-    const saved = await this.dagRepository.save(cherryPick);
-    if (saved) {
-      this.logger.log(`Registered new DAG ${saved.hash}`);
+      this.txRepository,
+      this.logger
+    );
+    let newDag = await this.dagRepository.findOneBy({ hash: zeroX(hash) });
```
Malak67, line 52: This can be new or existing so might be good to rename it. Is dag used anywhere? We could use just dag or currentDag. you decide.

services/explorer-indexer/src/modules/dag/dag.service.ts
```diff
@@ -41,35 +43,67 @@ export default class DagService {
...
+      dag.hash = zeroX(updateData.block);
+    }
     dag.pbftPeriod = parseInt(updateData.period, 10);
-    const updated = await this.dagRepository.save(dag);
-    if (updated) this.logger.log(`DAG ${updateData.block} finalized`);
+    try {
+      const updated = await this.dagRepository.save(dag);
+      if (updated) this.logger.log(`DAG ${updateData.block} finalized`);
+    } catch (error) {
+      console.error('NewDagBlockFinalizedResponse');
```
Malak67, line 105: Can you add both in a single console.error?  console.error('NewDagBlockFinalizedResponse', error)

services/explorer-indexer/src/utils/finder.ts
```diff
@@ -0,0 +1,44 @@
+import { Logger } from '@nestjs/common';
+import TransactionEntity from 'src/modules/transaction/transaction.entity';
+import { Repository } from 'typeorm';
+import util from 'util';
+
+export const safeSaveTx = async (
```
Malak67, line 6: I understand why you added this method here, but don't think it is a good place for it. Also the name of the file is not helpful. I think that any changes to the entity should be inside the service which is using the repository. That is more or less the nest way. 
So, I would move this function inside transactions.service and use it from there.

services/explorer-indexer/src/utils/finder.ts
```diff
@@ -0,0 +1,44 @@
...
+      tx = await repository.save(tx);
+    } catch (error) {
+      logger.warn(error);
+      logger.warn(`Skipping saving transaction: ${util.inspect(tx)}`);
+    }
+  }
+  return tx;
+};
+
+export const findTransactionsByHashesOrFill = async (
```
Malak67, line 32: Same comment as the previous one.

Malak67 (approved):

PR #268, opened 2022-10-11
Feature/add hisotrical sync

VargaElod23:


services/explorer-indexer/src/modules/node/historicalSyncer.service.ts
```diff
@@ -0,0 +1,256 @@
...
+
+      await this.pbftService.safeSavePbft(block);
+
+      // update sync state
+      this.syncState.number = block.number;
+      this.syncState.genesis = this.chainState.genesis;
+      this.syncState.hash = block.hash;
+
+      const completed = new Date();
+      console.log(
```
Malak67, line 244: Shouldn't we add logger here as well instead of console.log?

VargaElod23: Yes, it needs to be siwtched to logger everywhere, I just need console so that it prints during debug

PR #279, opened 2022-10-21
feat(explorer-shared): Moved entities and migrations to shared project

Malak67:
This solution moves all entities and migrations into the shared package.
Hence, when starting the indexer app, this needs to be started from the root via this command:
`yarn indexer-start`
This will first run any migration found inside the shared package and then starts the indexer app.


VargaElod23:
With this, when creating the according Dockerfiles for any package that uses the shared one, the building scope needs to be 
 `taraxa-platform root`. Just a note

packages/explorer-shared/tsconfig.json
```diff
@@ -16,7 +20,11 @@
     "module": "commonjs",
     "resolveJsonModule": true,
     "esModuleInterop": true,
-    "forceConsistentCasingInFileNames": true
+    "strictPropertyInitialization": false,
+    "forceConsistentCasingInFileNames": true,
```
VargaElod23, line 24: Hmm why was this necessary to define?

VargaElod23 (approved):
We should do an NPM release for this as adding this change makes management a bit more complex. To avoid having all of this into consideration when building packages that depend on it, it would be great if the package could be directly pulled from yarn or npm.

packages/explorer-shared/tsconfig.json
Malak67: I think it was there initially anway. It usually added by default from NestJs when creating a new project so I thought this project should have a similar tsconfig as the projects using NestJs.

VargaElod23: OKay. I'm fine with that

PR #287, opened 2022-10-25
Improvement/refactor arch to event driven

VargaElod23:


services/explorer-indexer/src/modules/syncer/syncer.module.ts
```diff
@@ -24,6 +25,17 @@ import { ConnectorsModule } from '../connectors';
         };
       },
     }),
+    BullModule.registerQueue(
+      {
+        name: 'pbfts',
```
Malak67, line 30: This will be the historical sync of the pbfts, right? We should add historical-pbfts as name. Same for dags.

Malak67 (approved):

PR #289, opened 2022-10-28
Hotfix/fix historical sync trigger

VargaElod23:


services/explorer-indexer/src/modules/pbft/pbft.service.ts
```diff
@@ -34,6 +34,15 @@ export default class PbftService {
     this.isRedisConnected = state;
   }
 
+  public async getSavedPbftPeriods() {
+    return (
+      (await this.pbftRepository
```
Malak67, line 39: I would do this using typeorm methods like .find and .select as it is more readable.

services/explorer-indexer/src/modules/pbft/pbft.service.ts
```diff
@@ -121,18 +130,31 @@ export default class PbftService {
...
+        for (const pbft of pbfts) {
           await this.dagService.findAndRemoveDagsForPbftPeriod(pbft.number);
-        })
-      );
-
-      this.logger.debug(
-        `Deleting ${pbfts?.length} PBFTs with number greater or equal than ${pbftNumber}`
-      );
+          if (pbft.transactions && pbft.transactions?.length > 0) {
+            await this.txService.deleteTransactions(
```
Malak67, line 137: Are you sure this won't crash anything? In this step the transactions are associated with Dags. That's why they are deleted when Dags are deleted. This might trigger a pg error, either at this step or when trying to delete and cascade Dags. We need to check at least.

Malak67 (approved):

Malak67:
> @rattrap So with this PR, the concurrent worker creation is handled. However, some edge cases need to be considered:
> 
> * Tho `bull` is good, there are some edge cases where the same block ends up with two workers. In this scenario, the underlying driver fails so we need to restart. @Malak67 will take a look at it whether it can be solved with some hack on the `typeorm` side
> 
> image
> 
> * In any other case we just put the `pbftPeriod` back into Redis and work on it later.

Ok so I added some fixes and it seems that some issues are fixed now. I don't get the Foreign Key error when deleting now, but I am not 100% sure it deletes anymore (although yesterday it was deleting).
The second issue when we run two nodejs processes and they both are trying to save the same PBFT, actually one has saved it and the other one is trying to save it again. This happens rarely and the only problem I see is that the process gets stucked. Because the pbft was saved by the other process. I updated the try, catch block and now it should pass this error. We need to double-check it.

PR #312, opened 2022-12-14
Fix/lower cpu demands

VargaElod23:
- [x] Fixes the heap mem usage.
- [x] Fixes the temp_file_limit issue.

### DIFF
--- a/services/explorer-indexer/src/modules/health/health.controller.ts
+++ b/services/explorer-indexer/src/modules/health/health.controller.ts
@@ -6,7 +6,7 @@ import {
   DiskHealthIndicator,
   MemoryHealthIndicator,
 } from '@nestjs/terminus';
-import { SyncerHealthIndicator } from './SyncerHealthIndicator';
+import { SyncerHealthIndicator } from './syncerHealthIndicator';
 
 @Controller('health')
 export class HealthController {
--- a/services/explorer-indexer/src/modules/historical-sync/historical.syncer.service.ts
+++ b/services/explorer-indexer/src/modules/historical-sync/historical.syncer.service.ts
@@ -192,7 +192,7 @@ export default class HistoricalSyncService implements OnModuleInit {
     // @note : Right now there is no way to get the genesis transactions to set:
     // Initial validators, initial balances for delegators and faucet.
     // big TODO
-    const chuncks = [
+    const chunks = [
       {
         name: QueueJobs.NEW_PBFT_BLOCKS,
         data: {
@@ -205,7 +205,7 @@ export default class HistoricalSyncService implements OnModuleInit {
 
     for (let i = 1; i <= maxVal; i++) {
       if (foundBlockNumbers.indexOf(i) == -1) {
-        chuncks.push({
+        chunks.push({
           name: QueueJobs.NEW_PBFT_BLOCKS,
           data: {
             pbftPeriod: i,
@@ -219,21 +219,28 @@ export default class HistoricalSyncService implements OnModuleInit {
       this.syncState.number < this.chainState.number;
       this.syncState.number++
     ) {
-      chuncks.push({
+      if (chunks.length === 1000) {
+        try {
+          const added = await this.pbftsQueue.addBulk(chunks);
+          if (added) {
+            this.logger.log(
+              `Added ${chunks.length} PBFT periods in Queue up until ${
+                chunks[chunks.length - 1].data.pbftPeriod
+              }`
+            );
+            chunks.length = 0;
+          }
+        } catch (error) {
+          this.logger.error(error);
+        }
+      }
+      chunks.push({
         name: QueueJobs.NEW_PBFT_BLOCKS,
         data: {
           pbftPeriod: this.syncState.number,
           type: SyncTypes.HISTORICAL,
         } as QueueData,
       });
-    }
-    try {
-      const added = await this.pbftsQueue.addBulk(chuncks);
-      if (added) {
-        this.logger.log(`Added ${chuncks.length} PBFT periods in Queue`);
-      }
-    } catch (error) {
-      this.logger.error(error);
     }
   }
 }
--- a/services/explorer-indexer/src/modules/historical-sync/historical.syncer.service.ts
+++ b/services/explorer-indexer/src/modules/historical-sync/historical.syncer.service.ts
@@ -1,7 +1,6 @@
 import { Injectable, Logger, OnModuleInit } from '@nestjs/common';
 import { ConfigService } from '@nestjs/config';
 import { zeroX } from '@taraxa_project/explorer-shared';
-import _, { add } from 'lodash';
 import { ChainState } from 'src/types/chainState';
 import DagService from '../dag/dag.service';
 import PbftService from '../pbft/pbft.service';
@@ -10,6 +9,8 @@ import { GraphQLConnectorService } from '../connectors';
 import { InjectQueue } from '@nestjs/bull';
 import { Queue } from 'bull';
 import { QueueData, QueueJobs, SyncTypes } from '../../types';
+import QueuePopulatorCache from './queuePopulatorCache';
+
 @Injectable()
 export default class HistoricalSyncService implements OnModuleInit {
   private readonly logger: Logger = new Logger(HistoricalSyncService.name);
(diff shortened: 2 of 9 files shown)


services/explorer-indexer/src/modules/health/health.controller.ts
```diff
@@ -6,7 +6,7 @@ import {
   DiskHealthIndicator,
   MemoryHealthIndicator,
 } from '@nestjs/terminus';
-import { SyncerHealthIndicator } from './SyncerHealthIndicator';
+import { SyncerHealthIndicator } from './syncerHealthIndicator';
```
rattrap, line 9: Why is this camel case ? The file name is in pascal case.

services/explorer-indexer/src/modules/historical-sync/historical.syncer.service.ts
```diff
@@ -219,21 +219,28 @@ export default class HistoricalSyncService implements OnModuleInit {
...
+      if (chunks.length === 1000) {
+        try {
+          const added = await this.pbftsQueue.addBulk(chunks);
+          if (added) {
+            this.logger.log(
+              `Added ${chunks.length} PBFT periods in Queue up until ${
+                chunks[chunks.length - 1].data.pbftPeriod
+              }`
+            );
+            chunks.length = 0;
```
rattrap, line 231: Whaaa' ? :D

services/explorer-indexer/src/modules/historical-sync/historical.syncer.service.ts
```diff
@@ -219,21 +219,28 @@ export default class HistoricalSyncService implements OnModuleInit {
       this.syncState.number < this.chainState.number;
       this.syncState.number++
     ) {
-      chuncks.push({
+      if (chunks.length === 1000) {
```
rattrap, line 222: Are you sure the problem comes from here and not this:
```
const foundBlockNumbers = await this.pbftService.getSavedPbftPeriods();
```

?

rattrap (changes requested):

services/explorer-indexer/src/modules/historical-sync/historical.syncer.service.ts
VargaElod23: Yes, I was able to reproduce the problem via my local setup by doing:

- Exported the dump from the QA database.
- Imported locally
- Set the 2Gig limit  in a local docker container

So then, the error we had was there and this one fixed it and I was running the producer constantly and rerunning it repeatedly for the latter part of the day.

However, that can also be optimized, I'll add a QB query for that too. Thanks for pointing it out 👍🏼

services/explorer-indexer/src/modules/historical-sync/historical.syncer.service.ts
VargaElod23: The problem was that there were too many jobs created all at once, that's why the heap got full, so now we create chunks of 1000, send it bulk to the queue and then clear the array. 😄

services/explorer-indexer/src/modules/health/health.controller.ts
VargaElod23: To make it uniform with the other namings

services/explorer-indexer/src/modules/historical-sync/historical.syncer.service.ts
VargaElod23: Okay so I optimized this one too to return only the `pbft.number` property and tested on 13047 PBFT entries:

- [ ] node --max-old-space-size=2048 dist/src/main
- [ ] node --max-old-space-size=1024 dist/src/main
- [ ] node --max-old-space-size=512 dist/src/main
- [ ] node --max-old-space-size=256 dist/src/main
- [ ] node --max-old-space-size=128 dist/src/main
- [ ] node --max-old-space-size=64 dist/src/main
 
At 64 MB it fails the chunk creation because of the heap memory, but we should be good to go when it comes to database reads as that passed even at 64MB

rattrap:
Some notes:

Find number of blocks saved:
```
SELECT COUNT("id") FROM "pbfts"
```

Find the last block number (should be number of blocks saved - 1):
```
SELECT MAX("number") FROM "pbfts"
```

Find missing block numbers:
```
SELECT s.i AS "missing_number"
FROM generate_series(0, <MAX_BLOCK_NUMBER>) s(i)
WHERE NOT EXISTS (SELECT 1 FROM "pbfts" WHERE "number" = s.i);
```

services/explorer-indexer/src/modules/historical-sync/historical.syncer.service.ts
rattrap: I think we can improve this further. 13k PBFT blocks is a small number.
The mainnet has 758k blocks after running for 1 month.

See comments on the PR.

rattrap (changes requested):

services/explorer-indexer/src/modules/pbft/pbft.service.ts
```diff
@@ -34,14 +34,20 @@ export default class PbftService {
...
 
-  public async getSavedPbftPeriods() {
-    return (
-      await this.pbftRepository.find({
-        select: {
-          number: true,
-        },
-      })
-    )?.map((entry) => entry.number);
+  public async getMissingPbftPeriods() {
```
VargaElod23, line 37: Hi @rattrap . So, I went on and changed the logic a bit so that the calculation is entirely delegated to the database, which is a major improvement. Thanks for the series generation example, it was a really good source of inspiration. I took it one step forward and tweaked it so that it returns all the `pbftPeriods` missing.

services/explorer-indexer/src/modules/historical-sync/historical.syncer.service.ts
```diff
@@ -192,48 +217,59 @@ export default class HistoricalSyncService implements OnModuleInit {
...
-    for (let i = 1; i <= maxVal; i++) {
-      if (foundBlockNumbers.indexOf(i) == -1) {
-        chuncks.push({
-          name: QueueJobs.NEW_PBFT_BLOCKS,
-          data: {
-            pbftPeriod: i,
-            type: SyncTypes.HISTORICAL,
-          } as QueueData,
-        });
+    const missingBlockNumbers = await this.pbftService.getMissingPbftPeriods();
```
VargaElod23, line 228: And now, instead of getting the ones saved and then checking them and not including those, we simply iterate through the ones we don't have a push them in with the same chunky approach.

services/explorer-indexer/src/modules/historical-sync/historical.syncer.service.ts
```diff
@@ -192,48 +217,59 @@ export default class HistoricalSyncService implements OnModuleInit {
...
     }
-    try {
-      const added = await this.pbftsQueue.addBulk(chuncks);
-      if (added) {
-        this.logger.log(`Added ${chuncks.length} PBFT periods in Queue`);
-      }
-    } catch (error) {
-      this.logger.error(error);
-    }
+    await this.pushChunksBulk(
```
VargaElod23, line 268: And, at last, to treat the scenario when there isn't 1000 periods, we need to `addBulk` one last time if there's any leftovers.

services/explorer-indexer/src/modules/pbft/pbft.service.ts
```diff
@@ -34,14 +34,20 @@ export default class PbftService {
...
+    const lastBlockSaved = (
+      await this.pbftRepository.query(
+        'SELECT MAX("number") as "last_saved" FROM public.pbfts'
+      )
+    )[0].last_saved;
+    const missingNumbers: number[] = (
+      await this.pbftRepository.query(
+        `SELECT s.i AS "missing_number" FROM generate_series(0, ${
+          lastBlockSaved || 0
+        }) s(i) WHERE NOT EXISTS (SELECT 1 FROM public.pbfts WHERE number = s.i)`
```
VargaElod23, line 47: Add table name from env

services/explorer-indexer/src/modules/pbft/pbft.service.ts
```diff
@@ -34,14 +34,20 @@ export default class PbftService {
...
-      await this.pbftRepository.find({
-        select: {
-          number: true,
-        },
-      })
-    )?.map((entry) => entry.number);
+  public async getMissingPbftPeriods() {
+    const lastBlockSaved = (
+      await this.pbftRepository.query(
+        'SELECT MAX("number") as "last_saved" FROM public.pbfts'
```
VargaElod23, line 40: add table name from env

rattrap (approved):

PR #313, opened 2022-12-14
Enhancement/query graph ql for claims

### LINKED ISSUE #309: [Community Site] Claim: Reduce Infura calls
There is a [piece of code](https://github.com/Taraxa-project/taraxa-platform/blob/develop/services/claim-backend/src/modules/claim/claim.service.ts#L334) in the Claim Service that does repeated Infura requests.

As of now, the data is indexed in TheGraph so we should redirect the request there.

- [ ] Write the graphQL query to fetch the same data.
- [ ] Remove the Infura request.

VargaElod23:
Closes #309 .


VargaElod23:
Tested with input :

```JavaScript
{
	"address": "0xdd144d2b6de680c9faf1ebf1b801f85ad4b4cfe3",
	"amount": "8000000000000000000000",
	"nonce":  "9256"
}
```

services/claim-backend/src/modules/claim/claim.service.ts
```diff
@@ -331,16 +336,20 @@ export class ClaimService {
...
         );
+        if (!indexedClaims)
+          throw new InternalServerErrorException(
+            `No such claim was indexed with nonce: ${nonce}`,
+          );
+        if (indexedClaims.length > 1)
+          throw new InternalServerErrorException(
+            `Multiple claims were indexed with nonce: ${nonce}`,
+          );
+        const confirmation = ethers.BigNumber.from(indexedClaims[0].amount);
```
rattrap, line 352: We shouldn't throw here. This is the account request.

services/claim-backend/config/reward.ts
```diff
@@ -3,4 +3,5 @@ import { registerAs } from '@nestjs/config';
 export default registerAs('reward', () => ({
   communitySiteApiUrl: process.env.COMMUNITY_SITE_API_URL,
   delegationApiUrl: process.env.DELEGATION_API_URL,
+  graphQLClaimUrl: `${process.env.GRAPH_QL_CLAIM_ENDPOINT}/subgraphs/id/${process.env.CLAIM_SUBGRAPH_ID}`,
```
rattrap, line 6: We already have a config value for this URL in `indexer.ts`

rattrap (changes requested):

services/claim-backend/src/modules/claim/claim.service.ts
VargaElod23:
```suggestion
```

services/claim-backend/config/reward.ts
VargaElod23: Yeah that's for the staking contract, this one is for the claim, but I'm moving this there.

rattrap: I mean let's also add this claim subgraph url in the indexer file.

rattrap: And it can include the subgraph id too

rattrap: Oh. Yeah. Nevermind

services/claim-backend/src/modules/claim/claim.module.ts
```diff
@@ -30,15 +32,23 @@ import { BlockchainModule } from '@taraxa-claim/blockchain';
...
   controllers: [
     BatchController,
     RewardController,
     AccountController,
     ClaimController,
   ],
-  providers: [ClaimService],
-  exports: [ClaimService],
+  providers: [ClaimService, GraphQLService],
+  exports: [ClaimService, GraphQLService],
```
rattrap, line 52: No need to export it. We're using it in the same module

rattrap (approved):

services/claim-backend/src/modules/claim/claim.module.ts
VargaElod23: Removed

PR #314, opened 2022-12-15
Feature/implement on chain undelegataion

### LINKED ISSUE #308: [Community Site] Un-delegate - new economics: Fix undelegation flow
After the on-chain economics update the undelegation flow doesn't work anymore. We need to find a way to fix that.
Discussion and details: https://discord.com/channels/419749122556297216/928006092124983366/1051445992370745375

VargaElod23:
Closes #308 .

### DIFF
--- a/services/delegation/src/modules/blockchain/blockchain.service.ts
+++ b/services/delegation/src/modules/blockchain/blockchain.service.ts
@@ -126,15 +126,11 @@ export class BlockchainService {
 
   async undelegate(address: string, amount: ethers.BigNumber) {
     try {
-      const tx = await this.contract.undelegate(address, {
+      const tx = await this.contract.undelegate(address, amount, {
         gasPrice: this.provider.getGasPrice(),
-        value: amount,
       });
-      await tx.wait();
-      if (tx) {
-        const receipt = await this.provider.getTransactionReceipt(tx.hash);
-        return receipt.blockNumber;
-      }
+      const receipt = await tx.wait();
+      return receipt.blockNumber;
     } catch (e) {
       console.error(
         `Could not undelegate ${amount.toString()} TARA to validator ${address}`,
@@ -145,14 +141,11 @@ export class BlockchainService {
 
   async confirmUndelegate(address: string) {
     try {
-      const tx = await this.contract.undelegate(address, {
+      const tx = await this.contract.confirmUndelegate(address, {
         gasPrice: this.provider.getGasPrice(),
       });
-      await tx.wait();
-      if (tx) {
-        const receipt = await this.provider.getTransactionReceipt(tx.hash);
-        return receipt.blockNumber;
-      }
+      const receipt = await tx.wait();
+      return receipt.blockNumber;
     } catch (e) {
       console.error(
         `Could not confirm undelegation for validator ${address}`,
--- a/services/delegation/src/modules/blockchain/blockchain.service.ts
+++ b/services/delegation/src/modules/blockchain/blockchain.service.ts
@@ -1,5 +1,5 @@
 import * as ethers from 'ethers';
-import { Injectable } from '@nestjs/common';
+import { Injectable, InternalServerErrorException } from '@nestjs/common';
 
 @Injectable()
 export class BlockchainService {
@@ -24,6 +24,9 @@ export class BlockchainService {
       '0x00000000000000000000000000000000000000fe',
       [
         'function delegate(address validator) payable',
+        'function undelegate(address validator, uint256 amount) external',
+        'function confirmUndelegate(address validator) external',
+        'function cancelUndelegate(address validator) external',
         'function getValidator(address validator) view returns (tuple(uint256 total_stake, uint256 commission_reward, uint16 commission, uint64 last_commission_change, address owner, string description, string endpoint) validator_info)',
         'function getValidators(uint32 batch) view returns (tuple(address account, tuple(uint256 total_stake, uint256 commission_reward, uint16 commission, uint64 last_commission_change, address owner, string description, string endpoint) info)[] validators, bool end)',
         'function registerValidator(address validator, bytes proof, bytes vrf_key, uint16 commission, string description, string endpoint) payable',
@@ -44,6 +47,17 @@ export class BlockchainService {
       defaultDelegationAmount,
       ownNodes,
     );
+  }
+
+  async getCurrentBlockNumber() {
+    try {
+      return await this.provider.getBlockNumber();
+    } catch (error) {
+      throw new InternalServerErrorException(
+        `Cound not collect current block number`,
+        error,
+      );
+    }
   }
 
   async getValidator(address: string) {
@@ -108,6 +122,43 @@ export class BlockchainService {
     }
 
     return false;
+  }
+
+  async undelegate(address: string, amount: ethers.BigNumber) {
+    try {
+      const tx = await this.contract.undelegate(address, {
+        gasPrice: this.provider.getGasPrice(),
+        value: amount,
+      });
+      await tx.wait();
+      if (tx) {
+        const receipt = await this.provider.getTransactionReceipt(tx.hash);
+        return receipt.blockNumber;
+      }
+    } catch (e) {
+      console.error(
+        `Could not undelegate ${amount.toString()} TARA to validator ${address}`,
+        e,
+      );
+    }
+  }
+
+  async confirmUndelegate(address: string) {
+    try {
+      const tx = await this.contract.undelegate(address, {
+        gasPrice: this.provider.getGasPrice(),
+      });
+      await tx.wait();
+      if (tx) {
+        const receipt = await this.provider.getTransactionReceipt(tx.hash);
+        return receipt.blockNumber;
+      }
+    } catch (e) {
+      console.error(
+        `Could not confirm undelegation for validator ${address}`,
+        e,
+      );
+    }
   }
 
   async rebalanceOwnNodes(addOneNode = false) {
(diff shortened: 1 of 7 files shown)


services/delegation/src/modules/blockchain/blockchain.service.ts
```diff
@@ -24,6 +24,9 @@ export class BlockchainService {
       '0x00000000000000000000000000000000000000fe',
       [
         'function delegate(address validator) payable',
+        'function undelegate(address validator, uint256 amount) external',
+        'function confirmUndelegate(address validator) external',
+        'function cancelUndelegate(address validator) external',
```
VargaElod23, line 29: @rattrap  I saw that there is this func too, shouldn't we use this and create another cron to clean up the "outdated" delegations (like the ones older than 5 days)?

services/delegation/src/modules/delegation/undelegation.entity.ts
```diff
@@ -0,0 +1,64 @@
...
+import { CreateUndelegationDto } from './dto/create-undelegation.dto';
+
+@Entity({
+  name: 'undelegations',
+})
+export class Undelegation {
+  @PrimaryGeneratedColumn()
+  id: number;
+
+  // @rattrap do we need this, and if so, where do we get it from?
```
VargaElod23, line 22: @rattrap FYI

services/delegation/src/modules/blockchain/blockchain.service.ts
rattrap: No. Only we do undelegations. And if we call `undelegate`, we want to confirm it, not cancel it.

services/delegation/src/modules/blockchain/blockchain.service.ts
```diff
@@ -110,6 +124,43 @@ export class BlockchainService {
     return false;
   }
 
+  async undelegate(address: string, amount: ethers.BigNumber) {
+    try {
+      const tx = await this.contract.undelegate(address, {
+        gasPrice: this.provider.getGasPrice(),
+        value: amount,
```
rattrap, line 131: `amount` is a param to the function, we don't want to send those tokens to the contract.

services/delegation/src/modules/blockchain/blockchain.service.ts
```diff
@@ -110,6 +124,43 @@ export class BlockchainService {
...
+      });
+      await tx.wait();
+      if (tx) {
+        const receipt = await this.provider.getTransactionReceipt(tx.hash);
```
rattrap, line 135: I think if you save the output of `await tx.wait();`, that should be the receipt. No need to call this again

services/delegation/src/modules/blockchain/blockchain.service.ts
```diff
@@ -110,6 +124,43 @@ export class BlockchainService {
...
+      console.error(
+        `Could not undelegate ${amount.toString()} TARA to validator ${address}`,
+        e,
+      );
+    }
+  }
+
+  async confirmUndelegate(address: string) {
+    try {
+      const tx = await this.contract.undelegate(address, {
```
rattrap, line 148: We wanna call confirm, not undelegate again.

services/delegation/src/modules/delegation/delegation-task.service.ts
```diff
@@ -52,4 +65,38 @@ export class DelegationTaskService implements OnModuleInit {
...
   }
+
+  @Cron('*/10 * * * *')
+  async confirmUndelegation() {
+    this.logger.debug('Starting undelegation confirmation worker...');
+    const currentTestnetBlock =
+      await this.testnetBlockchainService.getCurrentBlockNumber();
+    const currentMainnetBlock =
+      await this.mainnetBlockchainService.getCurrentBlockNumber();
+    if (currentMainnetBlock && currentTestnetBlock) {
```
rattrap, line 76: No need to make them dependant on one another.

services/delegation/src/modules/delegation/undelegation.entity.ts
rattrap: No. The user is always us.

services/delegation/src/modules/delegation/delegation-task.service.ts
```diff
@@ -52,4 +65,38 @@ export class DelegationTaskService implements OnModuleInit {
...
+            chain: NodeType.TESTNET,
+          },
+        });
+      const mainnetnetUndelegationsInScope =
+        await this.undelegationRepository.find({
+          where: {
+            confirmationBlock: LessThan(currentTestnetBlock),
+            chain: NodeType.MAINNET,
+          },
+        });
```
rattrap, line 90: We're selecting all undelegations even if they were already confirmed.

services/delegation/src/modules/delegation/delegation-task.service.ts
```diff
@@ -52,4 +65,38 @@ export class DelegationTaskService implements OnModuleInit {
...
+        await this.undelegationRepository.find({
+          where: {
+            confirmationBlock: LessThan(currentTestnetBlock),
+            chain: NodeType.TESTNET,
+          },
+        });
+      const mainnetnetUndelegationsInScope =
+        await this.undelegationRepository.find({
+          where: {
+            confirmationBlock: LessThan(currentTestnetBlock),
```
rattrap, line 87: `currentMainnetBlock`

services/delegation/src/modules/delegation/delegation.service.ts
```diff
@@ -590,4 +604,69 @@ export class DelegationService {
...
   }
+
+  private async undelegateFromChain(
+    nodeId: number,
+    type: string,
+    address: string,
+    amount: BigNumber,
+  ) {
+    let undelegationBlockNumber;
+    const targetChain =
```
rattrap, line 615: Don't see the point of this.

services/delegation/src/modules/delegation/delegation.service.ts
```diff
@@ -590,4 +604,69 @@ export class DelegationService {
...
+        amount,
+      );
+    }
+
+    if (type === NodeType.MAINNET) {
+      undelegationBlockNumber = await this.mainnetBlockchainService.undelegate(
+        address,
+        amount,
+      );
+    }
```
rattrap, line 629: No need to call undelegate. It will be called by the cron

services/delegation/src/modules/delegation/delegation.service.ts
```diff
@@ -590,4 +604,69 @@ export class DelegationService {
...
+      if (!savedUndelegation) {
+        throw new InternalServerErrorException(
+          `Could not save undelegation for ${address} with value ${amount} from node ${nodeId} on ${type}`,
+        );
+      }
+    }
+  }
+
+  public async confirmUndelegation(undelegation: Undelegation) {
+    if (undelegation.address) {
```
rattrap, line 653: When is it not set?

services/delegation/src/modules/delegation/delegation.service.ts
```diff
@@ -590,4 +604,69 @@ export class DelegationService {
...
+            undelegation.address,
+          );
+      } else if (undelegation.chain === NodeType.TESTNET) {
+        confirmationBlock =
+          await this.testnetBlockchainService.confirmUndelegate(
+            undelegation.address,
+          );
+      }
+      if (!confirmationBlock)
+        throw new InternalServerErrorException(
```
rattrap, line 667: This is a HTTP error, we're calling this from a cron.

services/delegation/src/modules/blockchain/blockchain.service.ts
VargaElod23: Yeah, that's right, but won't there be any "dangling" ones after some time?

services/delegation/src/modules/delegation/undelegation.entity.ts
VargaElod23: Cool.

services/delegation/src/modules/blockchain/blockchain.service.ts
VargaElod23: Yes, you're right, I missed this one.

services/delegation/src/modules/blockchain/blockchain.service.ts
VargaElod23: Yep.

services/delegation/src/modules/delegation/delegation.service.ts
VargaElod23: Didn't we agree upon creating these on the condition of `if (currentDelegation.gt(totalNodeDelegation))` at https://github.com/Taraxa-project/taraxa-platform/blob/cc79dc42d27e15e6df40cf5b34c74649584788f0/services/delegation/src/modules/delegation/delegation.service.ts#L345 and then let the corn handle only the confirmations?

services/delegation/src/modules/delegation/delegation.service.ts
VargaElod23: Should always be, its just a sanity check because contract calls will fail in case it isn't

services/delegation/src/modules/delegation/delegation.service.ts
VargaElod23: Right.

services/delegation/src/modules/delegation/delegation.service.ts
VargaElod23: We need a check later on, but I need to add this ti the ifs too so that it makes sense, yes.

services/delegation/src/modules/blockchain/blockchain.service.ts
VargaElod23: Right, corrected.

rattrap (approved):

PR #317, opened 2022-12-16
delegation: fixes changed addres balances

rattrap:
NOTE: When a user changes his address, the community rewards move from the old address to the new one.

### DIFF
--- a/services/claim-backend/src/modules/claim/claim.service.ts
+++ b/services/claim-backend/src/modules/claim/claim.service.ts
@@ -37,9 +37,15 @@ interface DelegationRewardResponse {
   amount: number;
 }
 
+enum RewardType {
+  COMMUNITY = 'COMMUNITY',
+  DELEGATION = 'DELEGATION',
+}
+
 interface Reward {
   address: string;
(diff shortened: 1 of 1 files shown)


services/claim-backend/src/modules/claim/claim.service.ts
```diff
@@ -79,27 +85,131 @@ export class ClaimService {
...
+
+        total = total.plus(this.floatToBn(reward.total));
+        claimable = claimable.plus(total);
+
+        if (reward.type === RewardType.COMMUNITY) {
+          communityTotal = communityTotal.plus(this.floatToBn(reward.total));
+        }
+
+        if (reward.type === RewardType.DELEGATION) {
+          delegationTotal = delegationTotal.plus(this.floatToBn(reward.total));
```
VargaElod23, line 130: Doesn't `ethers.BigNumber` treat floats too?  I guess if u implemented it then it doesn't but we should add a PR to them with this, definitely.

VargaElod23 (approved):
Small check on the sts

PR #319, opened 2022-12-20
feat(economics-rewards): Added script to read validators from chain

Malak67:


services/economics-rewards/.env.example
```diff
@@ -0,0 +1,8 @@
+PROVIDER=https://rpc.devnet.taraxa.io/
+DPOS_CONTRACT_ADDRESS=0x00000000000000000000000000000000000000fe
+
+DB_HOST=localhost
+DB_PORT=5433
+DB_USERNAME=postgres
+DB_PASSWORD=ixF6eor9escu
```
VargaElod23, line 7: 🤣  n1

VargaElod23 (changes requested):
Some small comments added, none critical

services/economics-rewards/src/index.ts
```diff
@@ -0,0 +1,146 @@
...
+import * as fs from 'fs';
+
+import { Contract, ContractInterface, ethers, utils } from 'ethers';
+import DposAbi from './abi/DposAbi.json';
+import * as dotenv from 'dotenv';
+import { Delegation, Reward, Validator, ValidatorData } from './types';
+import { initializeConnection, saveReward } from './database';
+dotenv.config();
+
+const getContractInstance = async () => {
```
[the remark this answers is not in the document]
Malak67: We don't have modules. It's just a couple of files. I thought of creating a class for it but it is way to simple to need that.

services/economics-rewards/src/index.ts
```diff
@@ -0,0 +1,146 @@
...
+async function main() {
+  initializeConnection();
+  const { currentBlock } = await getContractInstance();
+  const blocks = [];
+  for (let i = 1; i <= currentBlock; i++) {
+    console.log('Current block: ', i);
+    const validators = await getBlockValidators(i);
+    blocks.push(...validators);
+  }
+  fs.writeFile(`src/output/blocks.json`, JSON.stringify(blocks), 'utf8', () => {
```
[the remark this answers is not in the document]
Malak67: Yeah that's why we are saving it to a database and from there we can export it in a CSV. This json is just for testing. I will probably remove it afterwards.

services/economics-rewards/src/delegators.ts
```diff
@@ -0,0 +1,87 @@
+import { DelegationData, Delegator } from './types';
+import { getContractInstance } from './contract';
+import { saveDelegator } from './database';
+
+const getDelegations = async (
```
VargaElod23, line 5:
```suggestion
const getAndPersistDelegations = async (
```

Would be great to have it expressed in the name that the method not only returns them but saves them too.

services/economics-rewards/src/delegators.ts
```diff
@@ -0,0 +1,87 @@
...
+      const formatted = {
+        blockNumber: blockNumber,
+        blockTimestamp: timestamp,
+        blockHash: hash,
+        delegator: delegator,
+        validator: delegationData.account,
+        stake: delegationData?.delegation?.stake?.toString(),
+        rewards: delegationData?.delegation?.rewards?.toString(),
+      };
+      await saveDelegator(formatted);
```
VargaElod23, line 30:
```suggestion
      const savedDelegator = await saveDelegator(formatted);
      return savedDelegator as Delegator
```
Would be good to return the persisted value.

services/economics-rewards/src/delegators.ts
VargaElod23, line 5: What do you think about wrapping it into a `const` called `dposContractConnector` or something similar so that later on when called, it is clear that these methods fetch data from chain

services/economics-rewards/src/validators.ts
```diff
@@ -0,0 +1,50 @@
+import { Validator, ValidatorData } from './types';
+import { saveValidator } from './database';
+import { getContractInstance } from './contract';
+
+const getAllValidators = async (
```
VargaElod23, line 5: Same thing here with the wrapper

VargaElod23 (approved):
Small naming suggestions but overall LGTM 👍🏼

PR #327, opened 2023-01-03
Feature/explorer fix connection switch

VargaElod23:


services/explorer/src/index.tsx
```diff
@@ -3,22 +3,34 @@ import ReactDOM from 'react-dom/client';
...
+
 import {
   ExplorerNetworkProvider,
   ExplorerLoaderProvider,
   NodeStateProvider,
 } from './hooks';
+import { recreateAPIConnection, recreateGraphQLConnection } from './utils';
+import { TESTNET_API } from './api';
+
+export let API = TESTNET_API;
```
Malak67, line 22: I would move this back to types.ts or where we declare all the endpoints as we don't want to import it from here.

services/explorer/src/api/explorer-api/fetchDagsByAddress.ts
```diff
@@ -1,6 +1,6 @@
 import axios, { AxiosResponse } from 'axios';
 import { useQuery } from 'react-query';
-import { API } from '../types';
+import { API } from '../../index';
```
Malak67, line 3: This should be moved back (same for the other hooks).

services/explorer/src/utils/networks.ts
```diff
@@ -0,0 +1,57 @@
...
+  GRAPHQL_API_TESTNET,
+  GRAPHQL_API_DEVNET,
+  GRAPHQL_API_MAINNET,
+  MAINNET_API,
+  TESTNET_API,
+  DEVNET_API,
+} from '../api';
+
+export const recreateGraphQLConnection = (): string => {
+  const network = localStorage.getItem(SELECTED_NETWORK);
```
Malak67, line 12: We should not take the network directly from localStorage. There is a a hook useNetworkSelection which sets inside the context the current network (and it has a logic on deciding if it takes it first based on the curent domain or local storage and so on). So this method should receive the network as parameter.

services/explorer/src/utils/networks.ts
```diff
@@ -0,0 +1,57 @@
...
+    }
+    default: {
+      connectionString = GRAPHQL_API_TESTNET;
+      break;
+    }
+  }
+  return connectionString;
+};
+
+export const recreateAPIConnection = (): string => {
```
Malak67, line 35: This should also receive the network as parameter.

services/explorer/src/index.tsx
```diff
@@ -3,22 +3,34 @@ import ReactDOM from 'react-dom/client';
...
+export let graphQLClient = urqlCreateClient({
+  url: recreateGraphQLConnection(),
+});
 
-export const graphQLClient = urqlCreatClient({
-  url: GRAPHQL_API,
+window.addEventListener('storage', () => {
+  // When local storage changes, recreate the connections
+  graphQLClient = urqlCreateClient({
+    url: recreateGraphQLConnection(),
```
Malak67, line 31: React the currentNetwork from the context:
`const { currentNetwork } = useExplorerNetwork();`
And then pass it to the two methods.

Malak67 (changes requested):
Small changes needed

rattrap (changes requested):
Don't like a few things:
1. We know the URLs for mainnet, testnet and devnet. We don't need env vars for those. Let's just hardcode them like here - https://github.com/Taraxa-project/taraxa-platform/blob/feature/moves_mainnet_to_onchain_economics/services/community-site/src/utils/networks.ts#L15
2. Don't like the global variable thing. We have the network in the context, let's use the context
3. I think we can do better than reloading

Malak67 (approved):

rattrap (approved):

PR #328, opened 2023-01-05
fix(explorer): Fixing UI bugs

Malak67:


VargaElod23 (approved):
Tested, LGTM

services/explorer/src/components/Header/Header.effects.ts
```diff
@@ -55,7 +55,7 @@ export const useHeaderEffects = () => {
       number: searchBlockNumber,
       hash: searchHash,
     }),
-    pause: !searchBlockNumber && !searchHash,
+    // pause: !searchBlockNumber && !searchHash,
```
rattrap, line 58: Should these be commented?
If we don't need them let's just delete it

rattrap (approved):

services/explorer/src/components/Header/Header.effects.ts
Malak67: Removed

PR #334, opened 2023-01-10
index pbft hashes

VargaElod23:


packages/explorer-shared/src/entities/dag.entity.ts
```diff
@@ -19,6 +19,7 @@ export class DagEntity extends BaseEntity implements IDAG {
   }
 
   @PrimaryGeneratedColumn()
+  @Index()
```
VargaElod23, line 22: These have indexes by default

packages/explorer-shared/src/entities/pbft.entity.ts
```diff
@@ -18,10 +18,12 @@ export class PbftEntity extends BaseEntity implements IPBFT {
   }
 
   @PrimaryGeneratedColumn()
+  @Index()
```
VargaElod23, line 21: This is redundant too

packages/explorer-shared/src/entities/transaction.entity.ts
```diff
@@ -18,6 +18,7 @@ export class TransactionEntity extends BaseEntity implements ITransaction {
   }
 
   @PrimaryGeneratedColumn()
+  @Index()
```
VargaElod23, line 21: Same

Malak67 (approved):

PR #367, opened 2023-01-30
fix(community-backend): Moved community-backend outside services to legacy folder

Malak67:


rattrap:
We also need to modify the circleci workflow

rjonczy:
> We also need to modify the circleci workflow @rjonczy

Fixed.

PR #376, opened 2023-02-01
feat: Database restructure, removed prefix, new migration, custom ind…

Malak67:
…exes


VargaElod23 (approved):
LGTM

VargaElod23:
That will still result in the right name

PR #377, opened 2023-02-01
Feat/api/blocks improvement

Malak67:


VargaElod23 (approved):
LGTM, let's see it in action

PR #381, opened 2023-02-07
Enhancement/split tx consumer from pbft

VargaElod23:
### Current benchmark

- One tx consumer saves cca 100 TPM

### Changes

- PBFTs are registered instantly with empty tx-es
- DAGs are registered instantly with empty tx-es
- Removes log noise
- Adds job TTL specification directly to their addition
- DAGs now get put back to queue if upsert fails
- Handles correctly the PBFT update scenario


Malak67 (approved):
Let`s see how it`s performing

PR #382, opened 2023-02-08
Enhancement/split tx consumer from pbft

VargaElod23:


services/explorer-indexer/src/modules/live-sync/live.syncer.service.ts
```diff
@@ -42,10 +45,12 @@ export default class LiveSyncerService {
     private readonly dagService: DagService,
     private readonly pbftService: PbftService,
     private readonly txService: TransactionService,
-    private readonly graphQLConnector: GraphQLConnectorService
+    private readonly graphQLConnector: GraphQLConnectorService,
+    private readonly historicalSyncerService: HistoricalSyncService
   ) {
     this.logger.log('Starting NodeSyncronizer');
     this.isWsConnected = false;
+    this.queueCache = new QueuePopulatorCache(this.pbftsQueue, 10);
```
rattrap, line 53: If we use this in both modules, maybe we can move it to a separate, `common`, module

services/explorer-indexer/src/modules/live-sync/live.syncer.service.ts
```diff
@@ -167,15 +172,28 @@ export default class LiveSyncerService {
...
-          );
-          this.logger.debug(`Pushed ${formattedNumber} into PBFT sync queue`);
+          if (this.historicalSyncerService.isRunning) {
+            await this.queueCache.add({
+              name: QueueJobs.NEW_PBFT_BLOCKS,
+              data: {
+                pbftPeriod: formattedNumber,
+                type: SyncTypes.LIVE,
+              } as QueueData,
+              JobKeepAliveConfiguration,
```
rattrap, line 182: Seems like a duplicate block. We do these in both cases

rattrap (changes requested):

services/explorer-indexer/src/modules/live-sync/live.syncer.service.ts
VargaElod23: Yes, however, we do add the period to different places.

rattrap (approved):
LGTM

PR #390, opened 2023-02-24
Feature/add subgraph for dpos

VargaElod23:


Malak67 (dismissed):
LGTM

packages/dpos-subgraph/schema.graphql
```diff
@@ -0,0 +1,55 @@
...
+  delegator: String!
+  validator: String!
+}
+
+type CommissionChange @entity(immutable: true) {
+  id: ID!
+  commission: Int!
+  registrationBlock: Int!
+  applianceBlock: Int!
+  validator: String!
```
rattrap, line 54: Let's add the date here too.

rattrap (changes requested):
There's no need to store the undelegations. These can be queried from the contract

packages/dpos-subgraph/src/mapping.ts
```diff
@@ -0,0 +1,172 @@
...
+  const currentDelegationDataTo = getOrInitCurrentDelegation(to, delegator);
+  const currentDelegationDataFrom = getOrInitCurrentDelegation(from, delegator);
+
+  // Update delegations
+  updateOrAddDelegation(currentDelegationDataFrom, amount, false);
+  updateOrAddDelegation(currentDelegationDataTo, amount, true);
+}
+
+export function handleCommissionSet(event: CommissionSet): void {
+  const commission = event.params.commission;
```
rattrap, line 162: I think we can normalise the commission here to 0-100%

rattrap (changes requested):

packages/dpos-subgraph/schema.graphql
```diff
@@ -0,0 +1,55 @@
...
+#   commissionChanges: [CommissionChange!]! @derivedFrom(field: "validator")
+#   # isEligible: Boolean
+#   # info: BasicInfo!
+# }
+
+type Delegation @entity(immutable: false) {
+  id: ID!
+  amount: BigInt!
+  delegator: String!
+  validator: String!
```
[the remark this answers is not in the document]
rattrap: Yes. That's fine.

packages/dpos-subgraph/src/mapping.ts
rattrap: I understand. Let's keep it as is then.

packages/dpos-subgraph/schema.graphql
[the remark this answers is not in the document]
VargaElod23: Cool, I'll be adding it in that manner then

Malak67 (approved):

PR #406, opened 2023-03-15
Chore/code improv

Malak67:
- Splitted AddressInfo component into two separate components (AddressDetails and AddressTables)
- Adjusted Pagination types
- Removed conditional operator where it didn`t made sense
- Fixed container padding on small screens
- Updated node page to use indexer data for displaying weeks


VargaElod23 (approved):
LGTM 👍🏼

PR #427, opened 2023-03-27
Initial setup for testnet-delegation contract

Malak67:


packages/testnet-delegation/contracts/interfaces/IDelegation.sol
```diff
@@ -0,0 +1,30 @@
...
+interface IDelegation {
+    struct Validator {
+        string name;
+        address wallet;
+        bytes32 addressProof;
+        bytes32 vrfKey;
+        uint256 commission;
+        string ip;
+        uint256 stake;
+    }
```
rattrap, line 14: We don't need this anymore

packages/testnet-delegation/contracts/mocks/MockIDPOS.sol
```diff
@@ -0,0 +1,88 @@
+// (c) 2023-2024, Taraxa, Inc. All rights reserved.
+// SPDX-License-Identifier: MIT
+pragma solidity 0.8.18;
+
+interface MockIDPOS {
```
rattrap, line 5: Why do we have this file if we already have the `[IDPOS.sol](https://github.com/Taraxa-project/taraxa-platform/pull/427/files#diff-9e867ada1ee36ff0543f30e387f23be30403b1a164021467de2454a7d2104d5c)` one?

packages/testnet-delegation/contracts/DelegationOrchestrator.sol
```diff
@@ -0,0 +1,70 @@
...
+    constructor(address[] memory _internalValidators, address _dpos) payable {
+        internalValidators = _internalValidators;
+        dpos = IDPOS(_dpos);
+        for (uint256 i = 0; i < _internalValidators.length; ++i) {
+            validatorRegistered[_internalValidators[i]] = true;
+        }
+        require(
+            validatorRegistered[_internalValidators[_internalValidators.length - 1]] == true,
+            "Failed to set own validators"
+        );
```
rattrap, line 31: This won't ever happen

packages/testnet-delegation/contracts/DelegationOrchestrator.sol
```diff
@@ -0,0 +1,70 @@
...
+    }
+
+    function addInternalValidator(address _newValidator) external onlyOwner {
+        require(validatorRegistered[_newValidator] == false, "Validator already registered");
+        internalValidators.push(_newValidator);
+        require(
+            internalValidators[internalValidators.length - 1] == _newValidator,
+            "Failed to add validator to own validators array"
+        );
```
rattrap, line 40: Can't happen

packages/testnet-delegation/contracts/DelegationOrchestrator.sol
```diff
@@ -0,0 +1,70 @@
...
+import "./interfaces/IDelegation.sol";
+import "./interfaces/IDPOS.sol"; // import the DPOS contract interface
+import "@openzeppelin/contracts/security/ReentrancyGuard.sol";
+import "@openzeppelin/contracts/access/Ownable.sol";
+import "@openzeppelin/contracts/utils/Address.sol";
+import "@openzeppelin/contracts/security/Pausable.sol";
+
+contract DelegationOrchestrator is IDelegation, Ownable, Pausable, ReentrancyGuard {
+    using Address for address;
+    uint256 constant MIN_REGISTRATION_DELEGATION = 1000000000000000000000;
```
rattrap, line 14: You could add eth here

rattrap (changes requested):

rattrap (changes requested):
We should also check the internal validators balances before delegation to them. If they already have 2t+1, we don't need to delegate when registering new external validator

packages/testnet-delegation/contracts/mocks/MockIDPOS.sol
VargaElod23: Ah, I see now, this is only a part of the main interface with the necessary business logic for tests, that's why it was duplicated.

rattrap:
1. Simplify internal state (we only need an array that contains our validator node addresses and a mapping between owner -> validator)
2. No need to send tokens to the register function.

PR #432, opened 2023-03-28
Implemented wallet context for handling wallet notifications

Malak67:


services/community-site/src/pages/RunNode/Modal/EditValidator.tsx
```diff
@@ -50,21 +52,8 @@ const EditValidator = ({
...
-        setStep(2);
-        await res.wait();
-        setLoading(false);
-      } catch (e) {
-        // eslint-disable-next-line no-console
-        console.error(e);
-        setLoading(false);
-      }
+      onSuccess();
+      asyncCallback(setValidatorInfo, [validator.address, payload.description, payload.endpoint]);
```
rattrap, line 56: Why not giving it a function as a param? Seems cleaner and simpler

rattrap (changes requested):

services/community-site/src/pages/RunNode/Modal/EditValidator.tsx
Malak67: But I am. setValidatorInfo is the function, and the second param is the array of params the setValidatorInfo requires.

Malak67: Or maybe I don`t understand what you are suggesting

VargaElod23: I guess what Leo meant was a single parameter as such: `(args: any[]) => Promise<any>`

VargaElod23: With that, we could pass the methods we define on hooks directly into it like importing setValdiatorInfo() and then doing `asyncCallback(setValdiatorInfo([description, endpoint]))`

VargaElod23 (commented):
Left some clarifications, besides that it feels okay

services/community-site/src/pages/RunNode/Modal/EditValidator.tsx
Malak67: Got it, changed

VargaElod23 (approved):

PR #466, opened 2023-04-04
Allowing floating numbers when delegating

Malak67:


services/community-site/src/utils/compareDelegationTo.ts
```diff
@@ -0,0 +1,12 @@
+import { ethers } from 'ethers';
+
+export const compareDelegationTo = (delegation: string, to = '1000'): boolean => {
```
VargaElod23, line 3: Can we take this one abstraction level above and simply use it to compare a BigNumber to a float/Number ? This way we can use it to compare `event.target.value` in every sencario in both components

VargaElod23 (changes requested):
Small request added

services/community-site/src/utils/compareDelegationTo.ts
Malak67: Hmm I thought of doing something similar but I don't think we can use it in multiple places except these two where I am using it now. Because we compare it either with float or BigNumber, either lower or greater. So it can`t be that reusable. Also it is a headache to test each scenarion.

VargaElod23 (dismissed):

services/community-site/src/utils/compareDelegationTo.ts
```diff
@@ -0,0 +1,12 @@
+import { ethers } from 'ethers';
+
+export const compareDelegationTo = (delegation: string, to = '1000'): boolean => {
+  if (!delegation) {
+    return false;
+  }
+  const value = delegation || '0';
+  const [integerPart] = value.split('.');
+  const integerBigNumber = ethers.BigNumber.from(integerPart || '0');
+  const isLessThan1000 = integerBigNumber.lt(ethers.BigNumber.from(to));
```
rattrap, line 10: isLessThanTo or just return the value directly

rattrap (changes requested):
Also change the sidebar text

rattrap (approved):

PR #467, opened 2023-04-04
Removing Californicum from networks and replacing testnet with partic…

Malak67:
…ipate for community


VargaElod23 (dismissed):

rattrap (changes requested):
Also change the sidebar sectiom

rattrap (approved):

PR #468, opened 2023-04-04
Migrated node-status to typescript

Malak67:


VargaElod23 (approved):

services/node-status/tsconfig.json
```diff
@@ -0,0 +1,29 @@
+{
+  "compilerOptions": {
+    "target": "es5",
+    "downlevelIteration": true,
+    "lib": ["dom", "dom.iterable", "esnext"],
+    "allowJs": true,
+    "skipLibCheck": true,
```
VargaElod23, line 7: Would there be a prob if we'd remove the skipLibCheck?

VargaElod23 (approved):

services/node-status/tsconfig.json
Malak67: Not sure. NextJs adds it by default when creating a new typescript project

Malak67:
@rattrap We need to test this to see if it is still working as before.

services/node-status/package.json
```diff
@@ -1,19 +1,26 @@
...
+    "start": "next start",
+    "lint": "next lint"
+  },
+  "dependencies": {
+    "@taraxa_project/taraxa-ui": "^0.0.1",
+    "axios": "^0.26.1",
+    "ethers": "^5.6.0",
+    "@types/node": "18.15.11",
+    "@types/react": "18.0.33",
+    "@types/react-dom": "18.0.11",
```
rattrap, line 17: Should go in `devDependancies`

rattrap (changes requested):

rattrap (approved):

PR #480, opened 2023-04-11
Fix/delegation fixes

Malak67:


VargaElod23 (dismissed):

services/community-site/src/pages/Staking/Modal/Delegate.tsx
```diff
@@ -117,20 +118,21 @@ const Delegate = ({ balance, validator, ownDelegation, onSuccess, onFinish }: De
           fullWidth
           margin="normal"
           onChange={(event) => {
+            setDelegationTotal(event.target.value);
+          }}
+          onKeyUp={(event) => {
+            const inputValue = (event.target as HTMLInputElement).value;
             if (parseFloat(delegationTotal) > parseFloat(ethers.utils.formatEther(balance))) {
```
rattrap, line 125: Shouldn't we also check the inputValue here and not the delegationTotal?

services/community-site/src/pages/Staking/Delegation.tsx
```diff
@@ -121,39 +121,38 @@ const Delegation = ({ location }: { location: Location }) => {
       (async () => {
         setLoadingAccountData(true);
         setDelegations(await getDelegations(account));
+        setUndelegations(await getUndelegations(account));
         setLoadingAccountData(false);
```
rattrap, line 125: What happens if there's an error with one of the two statements above? I think it will load continuously

rattrap (changes requested):

services/community-site/src/pages/Staking/Delegation.tsx
Malak67: Yeah, correct. Better to keep it separate as it was.

services/community-site/src/pages/Staking/Modal/Delegate.tsx
Malak67: Yeah, missed it.

PR #490, opened 2023-04-12
Improved design for update validator page

Malak67:
image

image


VargaElod23 (approved):
Although it is a bit different than the mockup from Steven I like it and it is a bit more condensed. LGTM

Malak67:
> Although it is a bit different than the mockup from Steven I like it and it is a bit more condensed. LGTM

What is different? The big screen is the same as the mockup, the only difference is on smaller devices

VargaElod23:
AH yes, right. The first time I checked only the second image loaded(German internet speed 🐌 )

Malak67:
image

PR #491, opened 2023-04-13
Feat/table improvements

Malak67:
image

image

image


rattrap (changes requested):
1. Address before the name
2. Column name should be "Address / Nickname"

services/community-site/src/components/Nickname/Nickname.tsx
```diff
@@ -0,0 +1,18 @@
+import React from 'react';
+// import { formatValidatorName } from '../../utils/string';
+import './nickname.scss';
+
+interface NickNameProps {
+  address: string;
+  description?: string;
+}
+
+const NickName = ({ address, description }: NickNameProps) => {
```
rattrap, line 10: Should be called `Nickname`, just like the filename

services/community-site/src/interfaces/Validator.ts
```diff
@@ -15,6 +15,12 @@ export interface ContractValidator {
   info: ContractValidatorInfo;
 }
 
+export enum ValidatorStatus {
+  Grey = 'grey',
```
rattrap, line 19: Keys should be uppercase

services/community-site/src/interfaces/Validator.ts
```diff
@@ -15,6 +15,12 @@ export interface ContractValidator {
...
+  Yellow = 'yellow',
```
rattrap, line 20: Also don't like representing colors in an interface. Maybe we could call them:
- ELIGIBLE
- ELIGIBLE_INACTIVE
- NOT_ELIGIBLE

services/community-site/src/utils/getValidatorStatusTooltip.ts
```diff
@@ -0,0 +1,14 @@
+import { ValidatorStatus } from '../interfaces/Validator';
```
rattrap, line 1: This should be close to the enum, it's logic not an util

services/community-site/src/services/useExplorerStats.tsx
```diff
@@ -32,9 +34,18 @@ export default () => {
...
+          const validatorStatus = isEligible
+            ? producedBlocksInLast24hours
+              ? ValidatorStatus.Green
+              : ValidatorStatus.Yellow
+            : ValidatorStatus.Grey;
+
           return {
             ...validator,
-            isActive: diffHours < 24,
+            isActive: producedBlocksInLast24hours,
```
rattrap, line 47: Do we need this anymore?

Malak67: Yes, having this isActive boolean might be usefull. I noticed it used in 1 or 2 places so that`s why I left it.

services/community-site/src/pages/RunNode/Screen/updatevalidator.scss
```diff
@@ -0,0 +1,79 @@
+.updateValidatorForm {
```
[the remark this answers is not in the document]
Malak67: No, I mean we haven`t discussed changing the design for all forms, just for this page. We can make sure that the details pages look the same if that is what you mean.

services/community-site/src/components/Nickname/Nickname.tsx
```diff
@@ -0,0 +1,18 @@
...
+interface NickNameProps {
+  address: string;
+  description?: string;
+}
+
+const Nickname = ({ address, description }: NickNameProps) => {
+  return (
+    <div className="nicknameContainer">
+      <div className="address">{address}</div>
+      <div>{description}</div>
```
rattrap, line 14: If it's optional why do we print this empty div?

services/community-site/src/pages/RunNode/Table/TestnetValidatorRow.tsx
```diff
@@ -1,9 +1,10 @@
 import React from 'react';
```
rattrap, line 1: I don't think this is enough. We may not hit the testnet indexer for these nodes and I'm sure we don't have a `testnetDpos` contract instance that we can call `isEligible` on

rattrap (changes requested):

Malak67:
image

image

image

PR #493, opened 2023-04-17
Implements yield calculation and display

VargaElod23:
@rattrap in the RunNode page we have "Expected yield" in the table. DO we want to display the newly calculated yields or keep the MAX 20% there?

### DIFF
--- a/services/community-site/src/services/useExplorerStats.tsx
+++ b/services/community-site/src/services/useExplorerStats.tsx
@@ -1,6 +1,6 @@
 import { useCallback, useMemo } from 'react';
 
-import { Validator } from '../interfaces/Validator';
+import { Validator, ValidatorWithStats } from '../interfaces/Validator';
 import { networks } from '../utils/networks';
 import useApi from './useApi';
 import useMainnet from './useMainnet';
@@ -10,7 +10,7 @@ export default () => {
   const { chainId } = useMainnet();
 
   const updateValidatorsStats = useCallback(
-    async (validators: Validator[]) => {
+    async (validators: Validator[]): Promise<Validator[] | ValidatorWithStats[]> => {
       if (validators.length === 0) {
         return validators;
       }
@@ -24,7 +24,6 @@ export default () => {
           if (!stats.success) {
             return validator;
           }
-
           const lastBlockTimestamp = stats.response.lastPbftTimestamp;
           if (!lastBlockTimestamp) {
             return validator;
@@ -34,8 +33,9 @@ export default () => {
           const diffHours = diff / 1000 / 60 / 60;
           return {
             ...validator,
+            pbftsProduced: stats.response.pbftCount,
             isActive: diffHours < 24,
-          };
+          } as ValidatorWithStats;
         }),
       );
 
--- a/services/community-site/src/services/useExplorerStats.tsx
+++ b/services/community-site/src/services/useExplorerStats.tsx
@@ -9,13 +9,38 @@ export default () => {
   const { get } = useApi();
   const { chainId } = useMainnet();
 
+  const fetchValidatorStatsForWeek = useCallback(async (): Promise<
+    { address: string; pbftCount: number; rank: number }[]
+  > => {
+    let validatorStats: { address: string; pbftCount: number; rank: number }[] = [];
+    let start = 0;
+    let hasNextPage = true;
+    while (hasNextPage) {
+      try {
+        const allValidators = await get(
+          `${networks[chainId].indexerUrl}/validators?limit=100&start=${start}`,
+        );
+        if (allValidators.success) {
+          validatorStats = [...validatorStats, ...allValidators.response.data];
+          hasNextPage = allValidators.response.hasNext;
+          start += 100;
+        }
+      } catch (e) {
+        // eslint-disable-next-line no-console
+        console.error(e);
+        hasNextPage = false;
+      }
+    }
+    return validatorStats;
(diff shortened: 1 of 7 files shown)


rattrap:
We want to have this "Yield Efficiency" field in the "Run a node" page too

VargaElod23:
Sure, I'll add it there too rightaway

services/community-site/src/utils/validators.ts
```diff
@@ -0,0 +1,34 @@
+import { Validator, ValidatorWithStats, YieldedValidator } from '../interfaces/Validator';
+
+const calulcateValidatorYield = (validators: Validator[]): ValidatorWithStats[] | Validator[] => {
```
rattrap, line 3: Typo in function name

services/community-site/src/utils/validators.ts
```diff
@@ -0,0 +1,34 @@
+import { Validator, ValidatorWithStats, YieldedValidator } from '../interfaces/Validator';
```
rattrap, line 1: I see this as important business logic, not a "utility" function.

services/community-site/src/utils/validators.ts
```diff
@@ -0,0 +1,34 @@
...
+
+const calulcateValidatorYield = (validators: Validator[]): ValidatorWithStats[] | Validator[] => {
+  const hasStats = (validators[0] as ValidatorWithStats)?.pbftsProduced !== 0;
+  if (!hasStats) {
+    return validators;
+  }
```
rattrap, line 7: You are assuming that the first validator in the list will not have 0 blocks produced. You can use `every` or `some` here

services/community-site/src/utils/validators.ts
```diff
@@ -0,0 +1,34 @@
...
+  const validatorsWithStats = validators as ValidatorWithStats[];
```
rattrap, line 8: You're upcasting here. It's better to downcast.

services/community-site/src/utils/validators.ts
```diff
@@ -0,0 +1,34 @@
...
+  const validatorsWithStake: YieldedValidator[] = validatorsWithStats.map((v) => {
```
rattrap, line 9: I don't see the point of this `YieldedValidator` interface. The name doesn't even make sense because you added the "blocksPerStake" field, not the "yield"

services/community-site/src/utils/validators.ts
```diff
@@ -0,0 +1,34 @@
...
+    return {
+      ...v,
+      blocksPerStake: v.pbftsProduced || 0 / Number.parseFloat(v.delegation.toString()),
```
rattrap, line 12: This is hard to understand. If it's zero, let's just leave it as zero

services/community-site/src/utils/validators.ts
```diff
@@ -0,0 +1,34 @@
...
+    };
+  });
+  const minBlockRatio = validatorsWithStake.reduce((prev, curr) =>
+    prev.blocksPerStake! < curr.blocksPerStake! ? prev : curr,
+  );
+
+  const maxBlockRatio = validatorsWithStake.reduce((prev, curr) =>
+    prev.blocksPerStake! > curr.blocksPerStake! ? prev : curr,
+  );
```
rattrap, line 21: This could lead to errors. I would use `Math.min(...validatorsWithStake.map(v => v.blocksPerStake))`

services/community-site/src/utils/validators.ts
```diff
@@ -0,0 +1,34 @@
...
+  return validatorsWithStake.map((validator) => {
+    const yieldRatio =
+      ((parseFloat(validator.blocksPerStake.toString()) - minBlockRatio.blocksPerStake) /
```
rattrap, line 24: Si previously `validator.blocksPerStake` was a number 'cause you were comparing it and now it's not?

services/community-site/src/services/useExplorerStats.tsx
```diff
@@ -34,8 +33,9 @@ export default () => {
           const diffHours = diff / 1000 / 60 / 60;
           return {
             ...validator,
+            pbftsProduced: stats.response.pbftCount,
```
rattrap, line 36: This is the total pbft blocks produced. We need the number of blocks produced this week

rattrap (changes requested):

services/community-site/src/utils/validators.ts
VargaElod23: I added it inside utils because we don't have any other dedicated place for such business logic files inside the current folder tree as in the services we have only hooks. Should I create a new one and reorganize these?

services/community-site/src/utils/validators.ts
VargaElod23: Right, I wanted this to be a check of whether the attribute exists on the object, so an `undefined` check should've come here.

services/community-site/src/utils/validators.ts
VargaElod23: Yeah, that would be feasible, but then I'll need to change some of the business logic of the whole flow, starting with `useExplorerStats`, I'll do that for the next iteration

services/community-site/src/utils/validators.ts
VargaElod23: float parsing requires a string so we need to convert it before

rattrap (approved):

PR #512, opened 2023-05-19
Feat/taraxa UI improv

Malak67:
Reduced components complexity:

- Removed state from NetworkMenu
- Removed search component from header and added it in the explorer header
- Moved complex table from taraxa-ui to explorer and kept a simple version of it
- Removed Wallet page from community (as it is not used and not working)
- Removed MUI from explorer
- Removed MUI from community
- Moved table design to delegation file
- Removed unnecessary theming for taraxa-ui components (created Provider)


VargaElod23 (dismissed):
LGTM, maybe we wanna add a paginated table later on to the components(with fetch on step).

services/explorer/src/theme-provider/ExplorerThemeProvider.tsx
```diff
@@ -1,7 +1,6 @@
 import React, { FC } from 'react';
-import { ThemeProvider } from '@mui/material/styles';
-import { theme } from './theme';
 import { StylesProvider, createGenerateClassName } from '@mui/styles';
```
rattrap, line 2: Maybe we can also get rid of `mui/styles`

VargaElod23: @rattrap on #513 I was tinkering most of the morning to shadow it from @taraxa-ui, but it seems that typings are not portable because the `makeStyles` is dynamically defining the return type and we'd need to infer to it anyways via

```TypeScript
: keyof Props extends never
  ? // `makeStyles` where the passed `styles` do not depend on props
    (props?: any) => ClassNameMap<ClassKey>
  : // `makeStyles` where the passed `styles` do depend on props
    (props: Props) => ClassNameMap<ClassKey>;
```

I mean it works, but we lose the type definitions...

rattrap (approved):

PR #542, opened 2023-07-20
Feature/event logs

### LINKED ISSUE #531: [Explorer] Add support for event logs
Similar to https://etherscan.io/tx/0xe9c162c8b263e0188704d9f3790022c38b22418e3afc3275af7c19a099d064bb#eventlog

- New table in Transaction page (https://mainnet.explorer.taraxa.io/tx/0xf492750a5d51ff6a1a2566ac12133ae041486b6617d92c38646de9e6b22ba2f4) like the the other tables in the explorer (example https://mainnet.explorer.taraxa.io/block/0x8e085a8449585294136fe4aa4f65ddafa2347f65ebdbcba7e10d98099c11e38c)

## Data to Display:
1. Address
2. Name
3. Params

## API Call Requirements
- No pagination
- We don't do the request if the transaction is not a contract call of contract creation

## Response data layout
https://github.com/Taraxa-project/taraxa-indexer/pull/97#issue-1793488536

Event Logs

VargaElod23:
### Closes #531 .


Malak67 (dismissed):

services/explorer/src/api/types.ts
```diff
@@ -131,3 +131,21 @@ export interface TablePagination<T> {
 
 export type BlockTablePagination = TablePagination<BlockData>;
 export type TxTablePaginate = TablePagination<Transaction>;
+
+export type ToastData = {
+  display: boolean;
+  variant?: 'success' | 'error' | 'warning' | undefined;
+  text?: string;
+};
```
rattrap, line 139: Why would this be in the API?

rattrap (changes requested):

services/explorer/src/api/types.ts
VargaElod23: Since its being used in https://github.com/Taraxa-project/taraxa-platform/blob/68ae9a2e8e46cb78e607e77dd018d01f74646da4/services/explorer/src/api/sendRequestTokens.faucet.ts#L3 as callback type

PR #554, opened 2023-08-22
Feat/541/hex to dec

### LINKED ISSUE #541: [Explorer] Add switch to show transaction value in decimal or hex
image

## Requirements

Like on etherscan, we'd like to have the option to switch from hex to decimal on the transaction values.

Malak67:
Closes #541 

image
image
image
image


VargaElod23 (approved):
LGTM 👍🏼

services/explorer/src/pages/TransactionData/TopicDataDisplay.tsx
```diff
@@ -0,0 +1,71 @@
...
+
+  const handleClose = (val: string) => {
+    setValue(val);
+    setAnchorEl(null);
+    switch (val) {
+      case toggleValues[0].value:
+        setDecoded(hexValue);
+        break;
+      case toggleValues[1].value:
+        const addressHex = '0x' + hexValue.slice(-40);
```
VargaElod23, line 35: I just checked and this is the wrong way of doing it. This converts the address to decimal. However, on a log's topics we have only addresses, so all we need to do is to cut the padded zeroes to make the address readable.

services/explorer/src/pages/TransactionData/EventDataDisplay.tsx
```diff
@@ -0,0 +1,55 @@
...
+
+    setSelected(found.value);
+    switch (newValue) {
+      case toggleValues[0].value:
+        setDecoded(hexValue);
+        break;
+      case toggleValues[1].value:
+        const addressHex = '0x' + hexValue.slice(-40);
+        const addressDecimal = BigInt(addressHex).toString();
+        setDecoded(addressDecimal);
```
VargaElod23, line 32: image

This fails on empty data

VargaElod23 (changes requested):

services/explorer/src/pages/TransactionData/TopicDataDisplay.tsx
VargaElod23: Example: https://etherscan.io/tx/0x9051b74a7a2be1c071453f6b0ff076ab0173b07bb647927933c292c85a8f7076#eventlog

services/explorer/src/pages/TransactionData/EventDataDisplay.tsx
VargaElod23: Also, this won't work like this as we don't know the primitive that comes in here as base, so we should remove this. The main idea for this was to add a `decimal <> hex switch to the tx values and the block numbers` for debug purposes.

VargaElod23 (approved):

rattrap (approved):

PR #583, opened 2023-09-25
Added Dpos contract calls to new sdk package

### LINKED ISSUE #570: [SDK] taraxa-sdk package - add DPOS contract interface
By giving it a newtork ID or name, it should return an instance of the DPOS contract that will be used in the community site for read functions (get validator(s), getDelegations, etc.)

Malak67:
Closes #570

### DIFF
--- a/packages/taraxa-sdk/package.json
+++ b/packages/taraxa-sdk/package.json
@@ -18,7 +18,8 @@
   "author": "Taraxa",
   "peerDependencies": {
     "react": "^18.1.0",
-    "react-dom": "^18.1.0"
+    "react-dom": "^18.1.0",
+    "ethers": "^5.4.7"
   },
   "dependencies": {
     "@rollup/plugin-commonjs": "^17.0.0",
--- a/packages/taraxa-sdk/src/context/TaraxaDposClient.ts
+++ b/packages/taraxa-sdk/src/context/TaraxaDposClient.ts
@@ -1,0 +1,99 @@
+import { ethers } from 'ethers';
+import { Network, networks } from '../utils/networks';
+
+export class TaraxaDposClient {
+  public mainnetProvider: ethers.providers.JsonRpcProvider;
+
+  public browserProvider: ethers.providers.Web3Provider | null = null;
+
+  public signer: ethers.Signer | null = null;
+
+  public chainId: number;
+
+  public network: Network;
+
+  public mainnetDpos: ethers.Contract | null = null;
+
+  public browserDpos: ethers.Contract | null = null;
+
+  constructor(networkIdOrName: number | string) {
+    this.network = this.lookupNetwork(networkIdOrName);
+
+    if (!this.network) {
+      throw new Error('Invalid network ID or name');
+    }
+
+    this.mainnetProvider = new ethers.providers.JsonRpcProvider(
+      this.network.rpcUrl
+    );
+    this.chainId = this.getChainIdFromNetwork(this.network);
+
+    if (window.ethereum) {
+      this.browserProvider = new ethers.providers.Web3Provider(
+        window.ethereum,
+        'any'
+      );
+      this.signer = this.browserProvider.getSigner();
+    }
+
+    this.initializeContracts();
+  }
+
+  private initializeContracts() {
+    const abi = [
+      'function cancelUndelegate(address validator)',
+      'function claimCommissionRewards(address validator)',
+      'function claimRewards(address validator)',
+      'function confirmUndelegate(address validator)',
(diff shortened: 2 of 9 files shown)


VargaElod23 (commented):
Good start for a react scope but remember, we need this SDK to work not only on a frontend scope but for backend functionalities too like setting up automated stuff on a server that connects to the DPOS contract.

VargaElod23:
For now this is just a wrapper that provides configuration out of the box, but we need to think about what are the most use-cases and implement utilities for those. Can you take a look at some other SDKs a bit later on to see what they're doing in addition?

Malak67:
We are building this SDK step by step. So far it does what is described in the tasks. We will continue working on it and improving it. It is not the latest version.

packages/taraxa-sdk/examples/node.ts
```diff
@@ -0,0 +1,21 @@
+import { NetworkName } from '../src';
+import { DposClient } from '../src/DposClient';
+
+const runExample = async () => {
+  const networkName = NetworkName.MAINNET;
+  const privateKey =
+    'aae3250deafff925811a14af4d7d0c9ba88b234c9f98443f9a203231bd36e805'; // Replace with your private key
```
VargaElod23, line 7: Is this a legit privkey? Hope not

packages/taraxa-sdk/src/DposClient.ts
```diff
@@ -0,0 +1,296 @@
...
+  DPOS_CONTRACT_ADDRESS,
+  Network,
+  NetworkName,
+  ProviderType,
+  getNetwork,
+  getNetworkById,
+} from './networks';
+
+export class DposClient {
+  private _dpos: DposAbi;
```
VargaElod23, line 13:
```suggestion
  private dpos: DposAbi;

```
Naming with _ is a pattern in Js for private but since we're in Ts I don't think it is needed. However, it is up to your belief. I don't wanna nit on it

packages/taraxa-sdk/src/DposClient.ts
```diff
@@ -0,0 +1,296 @@
...
+    } else if (
+      providerType === ProviderType.WEB3 &&
+      typeof window !== 'undefined' &&
+      window.ethereum
+    ) {
+      provider = new ethers.providers.Web3Provider(window.ethereum, 'any');
+      this._signer = (provider as ethers.providers.Web3Provider).getSigner();
+    } else {
+      throw new Error('Invalid provider type or environment');
+    }
```
VargaElod23, line 85: This if else chain feels a bit too verbose, we could use a nice little switch case here(I switched to this in my personal projects cuz its soo easy to read compared to)

```suggestion
    const conditionKey = `${providerType}_${typeof window}`;

switch (conditionKey) {
  case `${ProviderType.RPC}_undefined`:
    provider = new ethers.providers.JsonRpcProvider(this.network?.rpcUrl);
    if (privateKey) {
      this._signer = new ethers.Wallet(privateKey, provider);
    }
    break;

  case `${ProviderType.WEB3}_object`:
    if (window.ethereum) {
      provider = new ethers.providers.Web3Provider(window.ethereum, 'any');
      this._signer = provider.getSigner();
    }
    break;

  default:
    throw new Error('Invalid provider type or environment');
}

```

packages/taraxa-sdk/src/DposClient.ts
```diff
@@ -0,0 +1,296 @@
...
+    return getNetwork(networkIdOrName);
+  }
+
+  async createNewTransactOpts(): Promise<ethers.Overrides> {
+    if (!this.signer) {
+      throw new Error('Signer is not initialized');
+    }
+
+    const address = await this.signer.getAddress();
+    const nonce = await this.provider.getTransactionCount(address, 'pending');
```
VargaElod23: Let's remove nonce calcs here

VargaElod23 (changes requested):

packages/taraxa-sdk/src/DposClient.ts
Malak67: I have a getter with that name that`s why I added the _

packages/taraxa-sdk/src/DposClient.ts
Malak67: It`s a lot more readable this way but it won`t handle the case where we are in an UI (React) app where we need an RPC connection but we also have window.ethereum.

VargaElod23 (approved):

rattrap (changes requested):

PR #584, opened 2023-09-29
Feat/yield history widget

### LINKED ISSUE #575: [Community Site] Yield History - Make the widget bigger
Current

Let's try to use the last 60 days worth of values and see how it looks. We should also move it to the center of the page and make it bigger.

Malak67:
Closes #575 
image


VargaElod23:
@reedvoid so I've been thinking about changing the block numbers and here are my concerns: 

- UNIX timestamps are too long, they look ugly on the chart
- We could do W1, W2, W3, etc. but then we still need a relative to which we display these Week numbers. Could be the start of year or smth, but we should align on this one

